PR #15, opened 2019-03-06
Add draft of Dask NEP-18 advances

pentschev:


pentschev:
@mrocklin let me know what you think of this.

mrocklin:
cc also @jakirkham

mrocklin:
I think it's a fine blogpost.  Some large scale comments/questions:

1.  You compare Dask+NumPy to CuPy.  I'm curious if it would make more sense to compare NumPy to CuPy
2.  You didn't mention the Scikit-Learn results.  I think that these are important to include, and also demonstrate honesty.  Your current numbers demonstrate that there is clear promise to this approach, which is all we really need to accomplish here.  This is clearly a work in progress.
3.  I think you're copying the format in http://blog.dask.org/2019/03/04/building-gpu-groupbys, which mostly just listed a bunch of issues.  FWIW I'm not sure that this is the most exciting way to start a post.  I was mostly just lazy this time around.  These posts may be better to pattern off of, they definitely got more readership
    - http://blog.dask.org/2019/01/03/dask-array-gpus-first-steps
    - http://blog.dask.org/2019/01/29/cudf-joins

    You should probably refer to the first one in your post regardless

    You might consider starting with the theme of "We want to run GLM algorithms on GPUs.  Can we reuse existing libraries to accomplish this quickly?" the answer being, "Yes, if we do a bit of work to align a few libraries together, but first, here are some results"

pentschev:
Thanks for the comments, replying:

1. That's a good point, I think it makes sense to compare also to NumPy alone, but I think Dask+NumPy should still be there, after all, that's the whole point of Dask-GLM, isn't it? Plus, I'm currently assuming (which may not be the case) that Dask+NumPy should be faster than NumPy alone.

1. That's another good point, I intentionally left those out for now because I didn't yet have the time to have a glance at what's going on. For example, I still don't know what solver Scikit-Learn is using by default, so that comparison may be unfair.

1. This was totally coincidental, I had already started writing this post in this format about 2 weeks ago. I agree to your point, but I think we're still too early for that kind of writing, there's not much exciting stuff I can write about with the current limitations, namely single-threaded Dask+CuPy and lack of more interesting Dask-GLM examples. I think if we wait 2-4 weeks we could do another much more interesting blog post in the format you're proposing, including what I mentioned in item 2. Regardless, I will refer to your previous Dask GPU post, thanks for reminding me.

mrocklin:
> Plus, I'm currently assuming (which may not be the case) that Dask+NumPy should be faster than NumPy alone.

Dask adds some overhead.  I think it's worth trying out NumPy in isolation as well.

> For example, I still don't know what solver Scikit-Learn is using by default, so that comparison may be unfair.

I don't think that users care.  All's fair in love, war, and performance.

Scikit-Learn uses solvers that are asymptotically better than those used in dask-glm.  Dask-GLM's solvers were designed to scale well, but they're not particularly good on a single node.

> there's not much exciting stuff I can write about with the current limitations

Well, there are larger scale things here that I think are very exciting.  You've gotten a pre-existing codebase to mostly work on GPUs without doing that much work.  From an ecosystem perspective that's pretty interesting.

pentschev:
> Scikit-Learn uses solvers that are asymptotically better than those used in dask-glm. Dask-GLM's solvers were designed to scale well, but they're not particularly good on a single node.

I didn't mean to say exclusively about compute performance, but just in general, perhaps quality of results could be very much different, and I was trying to avoid leaving an untruthful conclusion for the reader based on information I don't currently have, for example: "CuPy is slower than Scikit-Learn, thus it must mean it's quality is much higher".

> Well, there are larger scale things here that I think are very exciting. You've gotten a pre-existing codebase to mostly work on GPUs without doing that much work. From an ecosystem perspective that's pretty interesting.

Ok, so you're talking more from a design perspective now, indeed in that you're right. I was thinking more from performance and application points of view. I will then have to rethink a bit what and how to write it to emphasize that perspective.

mrocklin:
> I didn't mean to say exclusively about compute performance, but just in general, perhaps quality of results could be very much different, and I was trying to avoid leaving an untruthful conclusion for the reader based on information I don't currently have, for example: "CuPy is slower than Scikit-Learn, thus it must mean it's quality is much higher".

Sure, you can give the context above, saying that Scikit-Learn just uses better algorithms that converge more quickly .

You may also want to read through: http://blog.dask.org/2017/03/22/dask-glm-1

pentschev:
Alright, I updated this, please tell me what you think @mrocklin.

mrocklin:
Looking forward to reading through this.

Also, cc @jakirkham for review

_posts/2019-03-11-dask-nep18.md
```diff
@@ -0,0 +1,521 @@
...
+### Compute SVD on the NumPy array wrapped in Dask array
+
+First, of all, this is what you had to do _before_ the introduction of the
+`__array_function__` protocol:
+
+```python
+u, s, v = da.linalg.svd(dx)
+u.compute()
+s.compute()
+v.compute()
```
mrocklin: Note that we would actually call `u, s, v = dask.compute(u, s, v)` here in order to compute everything at once.

_posts/2019-03-11-dask-nep18.md
```diff
@@ -0,0 +1,521 @@
...
+```
+
+For which we get the following timing:
+
+```
+CPU times: user 8.97 s, sys: 653 ms, total: 9.62 s
+Wall time: 9.45 s
+```
+
+Giving us another 2x speedup over the single-threaded CuPy SVD computing.
```
mrocklin: I'm surprised by this.  I would not have expected much speedup here.  Do we know what happened?

_posts/2019-03-11-dask-nep18.md
```diff
@@ -0,0 +1,521 @@
...
+
+To conclude, we started from over 3 minutes and are now down to under 10
+seconds by simply dispatching the work on a different array.
```
mrocklin: Nice result!

_posts/2019-03-11-dask-nep18.md
```diff
@@ -0,0 +1,521 @@
...
+
+
+Application
+-----------
+
+We will now talk a bit about potential applications of the ``__array_function``
```
mrocklin:
```suggestion
We will now talk a bit about potential applications of the ``__array_function__``
```

_posts/2019-03-11-dask-nep18.md
```diff
@@ -0,0 +1,521 @@
...
+
+```
+# Fitting
+CPU times: user 9.66 ms, sys: 116 µs, total: 9.78 ms
+Wall time: 8.94 ms
+
+# Predicting
+CPU times: user 130 µs, sys: 327 µs, total: 457 µs
+Wall time: 1.06 ms
+```
```
mrocklin: Would it be possible to increase the size of the input a bit?  I think that dask-glm is only really relevant once we get to larger sizes.  

Also, I recommend using `LogisticRegression` rather than `LinearRegression` if that's an easy change to make.  I suspect that seeing something other than `LinearRegression` will be more motivating.

_posts/2019-03-11-dask-nep18.md
```diff
@@ -0,0 +1,521 @@
...
+### Performance for different Dask-GLM solvers
+
+To verify how Dask-GLM with Dask arrays would compare with CuPy arrays, we also
+did some linear regression benchmarking of Dask-GLM solvers. The results below
+were obtained from a training dataset with 1000 features of 100 dimensions,
+and 1000 test features.
+
+```
+Solver admm with Dask took 22444.307 ms to fit and 41.299 ms to predict
+Solver admm with CuPy took 263.742 ms to fit and 1.146 ms to predict
```
mrocklin: You might consider including a `max_iter=` keyword when solving.  According to @cicdw our stopping criteria in dask-glm need some work.

_posts/2019-03-11-dask-nep18.md
```diff
@@ -0,0 +1,521 @@
...
+Solver lbfgs with CuPy took 13.935 ms to fit and 0.260 ms to predict
+
+Solver newton with Dask took 785.150 ms to fit and 42.822 ms to predict
+Solver newton with CuPy took 22.480 ms to fit and 0.248 ms to predict
+
+Solver proximal_grad with Dask took 1902.620 ms to fit and 45.349 ms to predict
+Solver proximal_grad with CuPy took 11.700 ms to fit and 0.257 ms to predict
+
+Solver gradient_descent with Dask took 2483.603 ms to fit and 42.605 ms to predict
+Solver gradient_descent with CuPy took 14.580 ms to fit and 0.257 ms to predict
```
mrocklin: When you say "Dask" above you mean Dask+NumPy, correct?  Would it be hard to include straight NumPy numbers here as well?

mrocklin: You may also want to format this as a table.  I usually use HTML for this rather than markdown.  

https://github.com/dask/dask-blog/blob/gh-pages/_posts/2019-01-03-dask-array-gpus-first-steps.md#results

mrocklin (commented):
Some feedback.

In general this looks great to me.  The examples are very motivating and it's a fun read.

_posts/2019-03-11-dask-nep18.md
pentschev: > Would it be hard to include straight NumPy numbers here as well?

I suppose not, I will check that.

> You may also want to format this as a table. I usually use HTML for this rather than markdown.

Great idea, thanks for the suggestion!

_posts/2019-03-11-dask-nep18.md
pentschev: Yes, you're right, but in this case we're comparing only the backends, so it will probably not make a big difference, thought it would could make a bigger difference when comparing to scikit-learn. But for a scikit-learn comparison, it could seem like cheating, since we can't set the same parameter there.

_posts/2019-03-11-dask-nep18.md
pentschev: > Would it be possible to increase the size of the input a bit? I think that dask-glm is only really relevant once we get to larger sizes.

Unfortunately, not for fitting, as we run out of memory (I didn't want to dive deep now to check in detail why). I didn't try that for prediction though. I'll check.

> Also, I recommend using `LogisticRegression` rather than `LinearRegression` if that's an easy change to make. I suspect that seeing something other than `LinearRegression` will be more motivating.

I will check if it's possible without too big changes.

_posts/2019-03-11-dask-nep18.md
pentschev: To be honest, I didn't try to profile it at this time, but this is something we definitely should do in the near-future. I imagine the major reason is that with multithreading we help keeping the GPU busy, perhaps CuPy's implementation is not able to keep it busy during the whole computation. I know it uses cuSOLVER which is likely heavily optimized, but it's difficult to say if CuPy is actually making the best use of it without looking at the code.

Perhaps at some point https://github.com/dask/distributed/issues/2513 would help us with more profiling information. I hope I can soon take a look into that issue.

_posts/2019-03-11-dask-nep18.md
pentschev: Oh yes, that's a very good point, I will change that.

_posts/2019-03-11-dask-nep18.md
cicdw: Yea, stopping criteria for `proximal_grad` and `admm` are always finnicky; you might also want to include some sort of comparison between the actual outputs of the various model fits to ensure that they've all converged to something meaningful.

_posts/2019-03-11-dask-nep18.md
pentschev: Well, that's very unexpected. Using NumPy arrays is faster than CuPy, and _much_ faster than Dask:

```
Solver admm took 120.516 ms to fit and 0.492 ms to predict
Solver lbfgs took 13.858 ms to fit and 0.438 ms to predict
Solver newton took 52.420 ms to fit and 0.235 ms to predict
Solver proximal_grad took 4.047 ms to fit and 0.216 ms to predict
Solver gradient_descent took 6.207 ms to fit and 0.218 ms to predict
```

I could understand NumPy and CuPy being more or less close, but not Dask being absurdly slower than both. The code is basically this (plus imports and timing computation):

```python
X, y = make_classification()
lr.fit(X, y)
lr.predict(X)
```

_posts/2019-03-11-dask-nep18.md
stsievert: Huh, the classes in `dask_glm.linear_models` don't have an `n_iter_` parameter. https://github.com/dask/dask-glm/issues/77

_posts/2019-03-11-dask-nep18.md
pentschev: After a bit of profiling, this is what I found out for Dask-GLM with Dask array:

```
    14339    0.139    0.000    0.814    0.000 /home/pentschev/.local/lib/python3.5/site-packages/dask/local.py:430(fire_task)
    44898   19.945    0.000   19.945    0.000 {method 'acquire' of '_thread.lock' objects}
     4055    0.042    0.000   19.992    0.005 /usr/lib/python3.5/threading.py:261(wait)
    14339    0.107    0.000   20.234    0.001 /usr/lib/python3.5/queue.py:147(get)
    14339    0.018    0.000   20.253    0.001 /home/pentschev/.local/lib/python3.5/site-packages/dask/local.py:140(queue_get)
      122    0.117    0.001   22.327    0.183 /home/pentschev/.local/lib/python3.5/site-packages/dask/local.py:345(get_async)
      122    0.013    0.000   22.346    0.183 /home/pentschev/.local/lib/python3.5/site-packages/dask/threaded.py:33(get)
      122    0.004    0.000   22.733    0.186 /home/pentschev/.local/lib/python3.5/site-packages/dask/base.py:345(compute)
        1    0.020    0.020   23.224   23.224 /home/pentschev/.local/lib/python3.5/site-packages/dask_glm/algorithms.py:200(admm)
        1    0.000    0.000   23.267   23.267 /home/pentschev/.local/lib/python3.5/site-packages/dask_glm/utils.py:13(normalize_inputs)
        1    0.000    0.000   23.268   23.268 /home/pentschev/.local/lib/python3.5/site-packages/dask_glm/estimators.py:65(fit)
```

A big portion of the time seems to be spent on waiting for thread lock. Also, looking at the callers, we see 100 `compute()` calls departing from `admm()`, which means it's _not_ converging and stopping only at `max_iter` as @cicdw suggested:

```
/home/pentschev/.local/lib/python3.5/site-packages/dask/base.py:345(compute)                               <-     100    0.004   19.637  /home/pentschev/.local/lib/python3.5/site-packages/dask_glm/algorithms.py:197(admm)
```

Running with NumPy, the algorithm converges, showing only 7 `compute()` calls:

```
/home/pentschev/.local/lib/python3.5/site-packages/dask/base.py:345(compute)                          <-       7    0.000    0.120  /home/pentschev/.local/lib/python3.5/site-packages/dask_glm/algorithms.py:197(admm)
```

I'm running Dask 1.1.4 and Dask-GLM master branch, to ensure that my local changes aren't introduce any bugs. However, if I run my Dask-GLM branch and use CuPy as a backend, it also converges in 7 iterations.

To me this seems to suggest that we have one of those very well-hidden and difficult to track bugs in Dask. Before I spent hours with this, any suggestions what could we look for?

mrocklin: All multi-threaded code profiles as "waiting for thread lock" in the main thread.  It's hard to profile multi-threaded code in Python.  If you're going to profile I recommend using `.compute(scheduler="single-threaded")`

pentschev: I know it's harder to profile multithreaded code. But profiling's not the point from my last comment, it only helped me see that when we use Dask arrays, the algorithms don't converge, and that's definitely the main issue with it taking so much longer.

I'm now asking for ideas on obvious things I could look for, profiling is irrelevant if we really have a bug that needs to be fixed, and that's why I was pointing out that NumPy and CuPy both converge, but Dask does not.

Minimal code to reproduce:

```python
from dask_glm.estimators import LogisticRegression
from dask_glm.datasets import make_classification

X, y = make_classification()

lr = LogisticRegression(solver='admm')
lr.fit(X, y)
```

mrocklin: OK, we should fix that.  To be honest I'm not sure how at the moment.  

For the purposes of this blogpost perhaps it would be better to compare just numpy vs cupy, and leave dask out of it?

pentschev: And to run the code above on a NumPy array (that converges), it suffices to change the last line to:

```python
lr.fit(X.compute(), y.compute())
```

mrocklin: cc @cicdw if you have suggestions

pentschev: > OK, we should fix that. To be honest I'm not sure how at the moment.

Yes, that's alright, I was mainly interested if anybody knew of something obvious.

> For the purposes of this blogpost perhaps it would be better to compare just numpy vs cupy, and leave dask out of it?

Agreed. I will do that and mention we currently have a bug in Dask, that's why we're leaving it out.

mrocklin: > mention we currently have a bug in Dask

Just to be clear,my guess is that it's more likely a bug in dask-glm than dask itself.

pentschev: It could be too, but my first suspicion was Dask because both NumPy and CuPy converge, so Dask becomes the obvious suspect, which doesn't make it necessarily guilty.

pentschev: Also, Dask-GLM estimators were deprecated in https://github.com/dask/dask-glm/pull/66. Maybe they indeed have a bug which may or may not have been fixed in Dask-ML.

TomAugspurger:
I could be wrong, but I don't think the estimators have changed
substantially in dask-ml:
https://github.com/dask/dask-ml/commits/master/dask_ml/linear_model/glm.py

_posts/2019-03-11-dask-nep18.md
cicdw: I don't actually think this is a true bug anywhere; ADMM is implemented to work with potentially distributed data, and works by solving a sequence of subproblems on the different pieces of the data (and is only truly useful for _approximate_ large scale solutions).

I'm guessing that the partioning of your data when using dask.array is just not a good partioning.  (Relevant line of code: https://github.com/dask/dask-glm/blob/master/dask_glm/algorithms.py#L241)

pentschev: @cicdw you may be right, but then we would have a "bug" in the `make_classification()` or in the example provided in https://dask-glm.readthedocs.io/en/latest/estimators.html.

pentschev: I also didn't profile every single solver, but they all take considerably longer with Dask than NumPy, so possibly all of them aren't converging as fast as with NumPy arrays.

pentschev:
@TomAugspurger that's possible, I'll try to investigate causes later, but it's probably useful to try out those in Dask-ML too.

_posts/2019-03-11-dask-nep18.md
pentschev: The results from the new graphs using `LogisticRegression`.

_posts/2019-03-11-dask-nep18.md
pentschev: I'm marking this as resolved, since this arises from the (possible) bug we identified, plus the new graphs I added give us some nice visualization of the current state between NumPy and CuPy.

pentschev:
For the time being, I don't plan any more changes. Ready for another review.

mrocklin:
I think that this looks great.  The one suggestion I would make is in the axes of the timing plots.

The main point of these plots seems to be to compare the relative performance between NumPy and CuPy.  To this end, it would be useful to have the NumPy and CuPy performance numbers on the same axes, so that one can see visually when the CuPy numbers are higher or lower than NumPy's numbers.  Currrently they look visually similar at first until you look at the numbers on the y-axes.

I could imagine rearranging the plots so that there is a different plot per solver, throwing all of the solvers into one plot, or keeping the two separate plots, but fixing the axes of the second plot to match the axes of the first.

mrocklin:
(I'm also happy to just push the merge button, this is great as it is)

pentschev:
@mrocklin I think the split by solver is a better solution. When first plotting the graphs I thought it would be too cluttered, but I did it now and it looks fine to me. So if there's no more comments, from my side it's ok to merge.

mrocklin:
Those plots do look nice :)

mrocklin:
I plan to merge this PR in an hour if there are no objections.

mrocklin:
I've merged this in and it's live.  It's a bit late in the week though to start publicizing it.  I'll tweet about it early next week.

pentschev:
Now that you've mentioned it, maybe we should have waited until Monday to publish that. Well, too late now I guess. :)

mrocklin:
I've just made it a draft, which should keep it off of the TOC for now.

https://github.com/dask/dask-blog/commit/ca2fe51d84ffbc7f3a1eb5abbaaf06eed0414a60

pentschev:
Nice, please publish it at the most appropriate time, you probably have experience in knowing which time is best, and I don't.

jakirkham:
Thanks @pentschev. This is really great.

I read an earlier version of this and really like the improvements you have made since. @mrocklin and I played with [a benchmark based on your work]( https://github.com/mrocklin/dask-gpu-benchmarks/blob/3b4211d2b1ac59189b4cfd1f914009758aeaba68/cupy-svd.ipynb ) earlier this week and it really shows the improvements you have made here! 😄

Sorry I was late to review. Have submitted some small tweaks to the text in PR ( https://github.com/dask/dask-blog/pull/18 ). These are mostly minor syntax changes. Though please take a look and make sure I haven't lost something in the changes.

As another note, it would be nice to list the BLAS and LAPACK library used with NumPy for the CPU case (can just run `python -c "import numpy; numpy.show_config()"` to check). Would be good to get the version of the BLAS and LAPACK libraries as well.

PR #23, opened 2019-03-30
Add draft of numba stencil post

mrocklin:
cc @jakirkham @sklam @seibert for review (if you have time)


mrocklin:
Also @quasiben @pentschev you may find this interesting, just because it would be nice to eventually use GPU arrays here.

pentschev:
I already see a follow-up post to this one, doing the same operations in Numba+CUDA and generating some benchmarks too. Maybe this could be somewhere in a "Future Work" section, or perhaps somehow rephrased even under "What could be better". What do you think?

mrocklin:
I think that that would be a great idea.  Probably we couldn't use `numba.stencil` for a while, but using just normal [numba.cuda.jit](https://numba.pydata.org/numba-doc/dev/cuda/index.html) is probably pretty similar.

I think that it would be useful to see more examples of numba.cuda.jit on real-world problems regardless.  Combining those examples with Dask array sounds especially interesting.  

Presumably using the `numba.cuda.threadIdx` object would make the code somewhat similar to the stencil code anyway.  I am curious how well this can be coupled with `numba.guvectorize`.

pentschev:
I agree, I know `numba.cuda.jit` works with Dask distributed, so it would be very interesting to show that it works with some nice example that people can extend upon. I can try to come up with some ideas and write a blog post in the next couple of weeks.

mrocklin:
That would be great!

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+    '(n, m) -> (n, m)'
+)
+def smooth(x, out):
+    out[:] = _smooth(x)
+```
+
+This function knows that it consumes a 2d array of int8's and produces a 2d
+array of int8's of the same dimensions.
+
+This sort of annotation is a small change, but it lets other systems like Dask
```
jrbourbeau:
```suggestion
This sort of annotation is a small change, but it gives other systems like Dask
```

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+---------
+
+**Docs:** http://numba.pydata.org/
+
+When we run this function on a NumPy array, we find that it is slow, operating
+at Python speeds.
+
+```python
+x = np.ones((100, 100))
+timeit f(x)
```
jrbourbeau:
```suggestion
timeit _smooth(x)
```

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+def smooth(x, out):
+    out[:] = _smooth(x)
+```
+
+This function knows that it consumes a 2d array of int8's and produces a 2d
+array of int8's of the same dimensions.
+
+This sort of annotation is a small change, but it lets other systems like Dask
+enough information to use it intelligently.  Rather than call functions like
+map_blocks, we can just use the function directly, as if our Dask Array was
```
jrbourbeau:
```suggestion
`map_blocks`, we can just use the function directly, as if our Dask Array was
```

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+
+Or, they could use Dask array, which will handle both the pipelining and the
+parallelism (single machine or on a cluster) all while still looking mostly
+like a NumPy array.
+
+```python
+import dask_image
+x = dask_image.imread('/path/to/*.png')  # a large lazy array of all of our images
+y = x.map_blocks(smooth, dtype='int8')
+TODO: write
```
jrbourbeau:
```suggestion
```

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+When we run this function on a NumPy array, we find that it is slow, operating
+at Python speeds.
+
+```python
+x = np.ones((100, 100))
+timeit f(x)
+527 ms ± 44.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
+```
+
+But if JIT compile this function with Numba, then it runs more quickly.
```
jrbourbeau:
```suggestion
But if we JIT compile this function with Numba, then it runs more quickly.
```

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+Here is the full code, we'll dive into it piece by piece below.
+
+```python
+import numba
+
+@numba.stencil
+def _smooth(x):
+    return (x[-1, -1] + x[-1, 0] + x[-1, 1] +
+            x[ 0, -1] + x[ 0, 0] + x[ 0, 1] +
+            x[ 1, -1] + x[ 1, 0] + x[ 1, 1]) // 9
```
jakirkham: Would call this `mean_filter_2d` or similar. People in the image processing community will definitely know what this is. For everyone else they can easily search for more info with that name in hand (though a link to Wikipedia would be good too).

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+
+
+@numba.guvectorize(
+    [(numba.int8[:, :], numba.int8[:, :])],
+    '(n, m) -> (n, m)'
+)
+def smooth(x, out):
+    out[:] = _smooth(x)
```
jakirkham: Same as above

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+```python
+@numba.njit
+def smooth(x):
+    return _smooth(x)
+
+%timeit smooth(x)
+70.8 µs ± 6.38 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
+```
+
+For those counting, that's over 1000x faster.
```
jrbourbeau: Seems like an excitement-inducing speedup

```suggestion
For those counting, that's over 1000x faster!
```

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+    [(numba.int8[:, :], numba.int8[:, :])],
+    '(n, m) -> (n, m)'
+)
+def smooth(x, out):
+    out[:] = _smooth(x)
+
+
+# If you want fake data
+import dask.array as da
+x = da.ones((1000000, 1000, 1000), chunks=('auto', -1, -1), dtype='int8')
```
jakirkham: Would pick different test data that would be more interesting with a mean filter. Perhaps a checkboard pattern or something else.

jrbourbeau (commented):
Gave this a read through and left a few minor comments. Feel free to accept/ignore as you see fit. This is an interesting post that reads really well!

_posts/2019-03-30-numba-stencil.md
jakirkham: Might be worth comparing to equivalent functions from SciPy ndimage and/or scikit-image.

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+**NumPy Docs:** https://docs.scipy.org/doc/numpy-1.16.0/reference/c-api.generalized-ufuncs.html
+
+A generalized universal function (gufunc) is a normal function that has been
+annotated with typing and dimension information.  For example we can redefine
+our `smooth` function as a gufunc as follows:
+
+```python
+@numba.guvectorize(
+    [(numba.int8[:, :], numba.int8[:, :])],
+    '(n, m) -> (n, m)'
```
jakirkham: Would be better to make this floating point as this is a mean filter. Perhaps `float32`?

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+array of int8's of the same dimensions.
+
+This sort of annotation is a small change, but it lets other systems like Dask
+enough information to use it intelligently.  Rather than call functions like
+map_blocks, we can just use the function directly, as if our Dask Array was
+just a very large NumPy array.
+
+```python
+# Before gufuncs
+y = x.map_blocks(smooth, dtype='int8')
```
jakirkham: Might be worth mentioning the `mean_filter` function in `dask-image`.

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,323 @@
...
+    ```python
+    def smooth(x):
+        return _smooth(x)
+    ```
+
+    [numba/numba #3916](https://github.com/numba/numba/issues/3916)
+
+5.  The dask-image library could use a `imsave` function
+
+    [dask/dask-image #110](https://github.com/dask/dask-image/issues/110)
```
jakirkham: Does `store` or `to_zarr` not work? This is my sense of what people do today.

jakirkham:
Thanks for the write-up. Looks pretty good. Would suggest using floating precision and updating names to match common terminology. Also had some questions and suggestions about dask-image usage. Will try to give this another look after revisions.

_posts/2019-03-30-numba-stencil.md
mrocklin: I'm inclined to keep this first post very simple and big-picture-y.  To that end I'm inclined to stick with an array that I can construct in a single line.  Constructing datasets isn't what I want to highlight, so I want to avoid much code around it.

If there is a way to construct a good dataset in a one-liner then I'm all for it.  I don't mind having this example be completely trivial though.

I think that if we end up doing a follow up that actually uses real data, that would be great.

_posts/2019-03-30-numba-stencil.md
mrocklin: I think I'm going to target folks other than imaging scientists to start out.  For the broader population I think that `smooth` is probably more approachable and less jargon-y.  I think that it would be good though to follow up on this blogpost with an actual imaging problem.

Maybe this is something that you would be interested?  Then maybe you and @pentschev could follow that with a GPU'ified version?

_posts/2019-03-30-numba-stencil.md
mrocklin: For the future, it would be good to think of some array examples for the `dask.datasets` module.

_posts/2019-03-30-numba-stencil.md
mrocklin: Will do

_posts/2019-03-30-numba-stencil.md
mrocklin: Ah indeed, that probably would have been better.  At this point I've already mostly moved on from this post.  So I'm going to lump this into the "I'm not trying to make a real example" comments from before.   This would have been good though.

_posts/2019-03-30-numba-stencil.md
mrocklin: Actually, I'm not finding that function in dask-image

_posts/2019-03-30-numba-stencil.md
mrocklin: Do you happen to know the operation to do here?

_posts/2019-03-30-numba-stencil.md
pentschev: While I agree with @mrocklin's argument about targeting a broader audience, I think it's worth a mention in the text about @jakirkham's suggestion, something like "the smooth function above/below is commonly referred to as the 2D mean filter in the image processing community".

_posts/2019-03-30-numba-stencil.md
pentschev: I think there's no direct operation for a mean filter, but you can use `scipy.ndimage.filters.convolve`, with `weights=[[1/9.0]*3]*3`.

_posts/2019-03-30-numba-stencil.md
pentschev: I think it's `dask_image.ndfilters.uniform_filter`, with `size=3`.

_posts/2019-03-30-numba-stencil.md
pentschev: Actually, you could also use `scipy.ndimage.uniform_filter` with `size=3` here.

_posts/2019-03-30-numba-stencil.md
jakirkham: Ah sorry appears I have misremembered the name. Thanks @mrocklin for catching that and @pentschev for providing the correct function.

jakirkham:
This comes up in a few threads above. So am pulling the discussion out of the threads to make it easier to follow. Hope that is ok.

> I think I'm going to target folks other than imaging scientists to start out. For the broader population I think that smooth is probably more approachable and less jargon-y. I think that it would be good though to follow up on this blogpost with an actual imaging problem.

If that is the case, this is something I missed while reading it (and still find a bit confusing). There are a few references to working with image data and some specific mentions about adding desired functionality to dask-image. If the goal is to target a broader group outside of the image processing community, then I would be inclined to drop this sort of stuff from the blogpost (even though the extra exposure to dask-image is great!). This leaves something for others to follow-up on later and then this blogpost does a nice job of setting the stage, which would also be great.

_posts/2019-03-30-numba-stencil.md
mrocklin: Where would you recommend this note?  I briefly tried looking for a place but it always seemed to break the flow of the document.

_posts/2019-03-30-numba-stencil.md
mrocklin: Thanks!  They're about the same performance.  I've added a note.

_posts/2019-03-30-numba-stencil.md
mrocklin: Would you like for me to remove this comment?

_posts/2019-03-30-numba-stencil.md
```diff
@@ -0,0 +1,325 @@
...
+
+# If you have actual data
+import dask_image
+x = dask_image.imread.imread('/path/to/*.png')
+
+y = smooth(x)
+# dask.array<transpose, shape=(1000000, 1000, 1000), dtype=int8, chunksize=(125, 1000, 1000)>
+```
+
+Lets break this down a bit
```
pentschev:
```suggestion
Note: the `smooth` function above is more commonly referred to as the 2D mean filter in the image processing community.

Now, lets break this down a bit
```

pentschev: @mrocklin I thought of something like this. I think it's not of uttermost importance though, if you really hate the way it looks after this, I'm fine if you just skip it.

_posts/2019-03-30-numba-stencil.md
jakirkham: Yeah, having some functions for generating test data could be good. Shipping testing data might be doable if we constrain the size a bit.

For a checkerboard pattern, normally I do something like this.

```suggestion
import numpy as np

s = (1000000, 1000, 1001)
x = da.arange(np.prod(s), chunks='auto').reshape(s) % 2
```

<br>

Mainly suggested this because it makes the change apparent to the end user if they wish to play with the code. Though this is probably more important with an example and less important with a blogpost.

Edit: IOW it's not critical to include this change.

_posts/2019-03-30-numba-stencil.md
jakirkham: We can leave it if you prefer. More of a question really.

_posts/2019-03-30-numba-stencil.md
jakirkham: Sounds like this was resolved by Peter's suggestion.

mrocklin:
I think that everything is resolved.  I will plan to merge this tomorrow morning if there are no further comments.

pentschev:
@mrocklin just checking, did you forget to merge this or are you waiting on purpose?

mrocklin:
Thanks for the reminder @pentschev !

PR #24, opened 2019-04-12
Input dask array to np.linalg.svd

jrbourbeau:
I could be missing something, but are these two `np.linalg.svd` function calls supposed to have dask arrays as input?

cc @pentschev


pentschev:
Ah, yes, thanks for spotting that. Too bad you didn't review the original draft, you would have found it earlier. :)

LGTM.

jrbourbeau:
I'll make sure to review next time. Really nice post btw, I enjoyed reading it.

pentschev:
Very glad you like it!

PR #25, opened 2019-05-02
Update GPU Dask Arrays, first steps

pentschev:
@mrocklin I think this is an important update to reflect the current state of CUDA support for Dask. I'm not sure whether it's a good idea to simply update the post, but the post as it is now is misleading and not easily reproducible.


mrocklin:
Fine with me.  Merging.

Although in the future I recommend that we don't update old blogposts (the burden here is too high).  Instead I think that we should write a disclaimer at the top when a blogpost is on new and experimental software.

pentschev:
I'm fine with writing a disclaimer on what has been updated and why, but I think the burden of having (some, not all) outdated posts that are often used as reference for reproducing results is far higher. Would you like me to write a disclaimer at the top?

mrocklin:
I don't have a strong preference.

PR #29, opened 2019-06-03
Adding possible task-stream figures to dgx blog (draft)

rjzamora:
@mrocklin I added task-stream figures for the SVD benchmark.  Let me know if this/isn't what you were thinking.

Note that html figures probably need to be relocated


mrocklin:
Thanks @rjzamora !

I recommend the following changes:

1.  Include a minimal implementation setting up the dask array part of the problem.  This is probably pulling out a couple cells of the notebook and including them here.  I'm guessing the code that creates the random array, calls the svd computation, and then the persist call.
2.  Maybe don't mention the difference between LocalCUDACluster (which readers probably won't be familiar with) and DGX, and instead maybe just say "before UCX" and "after UCX"
3.  Add yourself in the `Author` value at the top of this post
4.  Move the plots here to https://github.com/mrocklin/raw-host (or some similar repository) that we can use to hold and server larger artifacts like this

PR #30, opened 2019-06-04
Initial blogpost about loading image data

### LINKED ISSUE #26: Blogpost on loading TIF data into a Dask array
It would be handy to have a blogpost about loading image data into Dask Array, and then possibly storing it into some other format, like HDF5 or Zarr. 

It might be nice to have both a trivial example here (a simple stack of images), as well as a less-trivial example (if those are common).

cc @jakirkham , perhaps the world's leading expert

jakirkham:
Provides a simple intro to users about how they can load image data with Dask to form a Dask Array. Caters to the common use case of a bunch of image files that tile a larger image volume with multiple channels potentially acquired over time.

Closes https://github.com/dask/dask-blog/issues/26


_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,150 @@
+---
+layout: widepost
```
mrocklin:
```suggestion
layout: post
```

I had to use widepost in the previous post just to get certain output to fit.  I think that the normal `post` layout usually looks better.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,150 @@
...
+
+Series Overview
+---------------
+
+A common case in fields that acquire large amounts of imaging data is to write
+out smaller acquisitions into many small files. These files can tile a larger
+space, subsample from a larger time period, and may contain multiple channels.
+The acquistion techniques themselves are often state of the art and constantly
+pushing the envelope in term of how large a field of view can be acquired, at
+what resolution, and what quality (also known as signal-to-noise ratio).
```
mrocklin:
```suggestion
what resolution, and what quality.
```

This is a strongly voiced first paragraph.  I recommend not ending on a parenthetical, just to keep with that strength.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,150 @@
...
+Motivation
+----------
+
+Loading large image data can be a complex and often unique problem. Different
+groups may choose to store this across many files on disk, a commodity or
+custom database solution, or they may opt to store it in the cloud. Not all
+datasets within the same group may be treated the same for a variety of
+reasons. In short, this means loading data is a hard and expensive problem.
+
+Despite data being stored in different ways, often group want to reapply the
```
mrocklin:
```suggestion
Despite data being stored in many different ways, often groups want to reapply the
```

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,150 @@
...
+steps, it may be very difficult if not impossible to reuse an existing
+pipeline. In other words, there is friction between the loading and analysis
+steps, which frustrates efforts to make things reusable.
+
+Having a modular and general way to load data makes it easy to present data
+stored differently in a standard way. Further having a standard way to present
+data to analytical pipelines allows that part of the pipeline to focus on what
+it does best, analysis! In general, this should decouple these to components in
+a way that improves the experience of users involved in all parts of the
+pipeline.
```
mrocklin: I think that we probably need to reduce the exposition leading this post.  There are seven paragraphs of solid text before we get to non-text details like code, a table of results, or an image.  My experience is that most readers have a shorter attention span than this.

Some ideas:

1.  Drop the Series Overview section for now.  We can add something later after we have a series.
2.  Cut verbosity

    > Before: This post explores a few simple workflows that a user might leverage to load
their image with Dask.

    > After: We load image data with Dask.  
3.  Include a teaser code sample pretty much immediately afterwards in a "this is where we're going" comment, just so readers have a reason to stick around.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+Loading image data with Dask
+----------------------------
+
+A pretty typical case to see is a large directory of image files. In particular
```
mrocklin: Totally subjective, but I enjoy removing filler text.  I'll provide this suggestion as an example to encourage thought, but it's not a big deal.

```suggestion
A typical case is a large directory of image files. In particular,
```

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,150 @@
...
+TIFF files are quite common, though other formats could show up as well. It is
+common to see a naming structure that mirrors that shown below. The `<i>`
+represent different numbers.
+
+```
+mydata_ch<i>_<j>t_<k>x_<l>y_<m>z.tif
+```
```
mrocklin: I would be interested to know what common meanings for these are, perhaps a list like 

-  x, y, and z coordinates in space
-  A coordinate in time
-  Sample number
-  Channel, like red, blue, green if operating with optical light
-  ...

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,150 @@
...
+
+To load a single one of these a user might right some code like this:
```
mrocklin:
```suggestion
To load a single image, a user might write some code like the following:
```

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+While this works fine for reading a single image into memory, this will quickly
+become prohibitive once we try to load more data than we have memory. So we
+might replace this with something like this.
+
+```python
+from dask.array.image import imread
+a = imread("mydata_ch<i>_<j>t_<k>x_<l>y_<m>z.tif")
+```
```
mrocklin: Thoughts on starting with dask delayed instead?

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,150 @@
...
+chunks we would need to iterate over our data some how and combine the chunks
+together. Suppose we have one dimension to combine chunks along, we might write
+some code that looks like that below. We would load all of the chunks of data
+we have join them together with Dask Array's `block` to form a larger array. We
+can easily extend this to more dimensions.
+
+```python
+data = []
+for fn in files:
+    data.append(dask.array.imread(fn))
```
mrocklin: Are we reading each image here to get the shape and dtype?  I would expect this to become prohibitively expensive.  

When I do this I tend to do the following:

```python
sample = skimage.io.imread(fn)

data = [dask.delayed(skimage.io.imread)(fn) for fn in filenames]
data = [da.from_array(x, shape=sample.shape, dtype=sample.dtype) for x in data]
```

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,150 @@
...
+data = dask.array.block(data)
```
mrocklin: I think that `dask.array.block` could use some explanation here.  We might also want to mention `dask.array.concatenate` and `dask.array.stack`.  This might also be a good opportunity to use the `_repr_html_` work, which, fortunately for us, generates HTML that we can plop into this post.

mrocklin: (this is probably my most major comment)

One way to structure this post might be as follows:

1.  We got a bunch of interesting looking data from a leading edge microscopist.  Look at this cool image of the data as rendered nicely in 3d (cool image from IndeX)
2.  But!  Before we can analyze this data, we have to get it out of a bunch of TIFF files.  Gross! (truncated output of calling `ls`)
3.  This is actually pretty typical (now that we've hooked the reader, we can add some exposition about imaging data and tiff files and why this is the state of things.  They got a little endorphin boost from the image so we can throw some text at them now)
4.  So lets load this data with NumPy for analysis.  We would normally use scikit-image
5.  But our data is too big!  (maybe some `du -hs` call here on a stack of real image data, like Gokul's)
6.  So we can use Dask arrays to wrap around many TIFF files
7.  Example with dask delayed and da.from_delayed
8.  Now we have this list of lazy single-chunked arrays.  How do we want to combine those together?  Look at da.concatenate and da.stack, perhaps using the `_repr_html_` output (which I think could be very neat!)
9.  But in our case it's actually way more complex than this, because we don't have just one axis along which we want to concatenate/stack things.  This brings us to `dask.array.block`, where we place our delayed images into an array.
10.  Our actual code here looked like this (copy from your existing notebook)
11.  This is kind of gross, but the grossness here reflects the grossness of the layout.  What's nice is that once we do this, we have a nice clean dask array to work from
12.  Now that we've done this, lets write out to an analysis ready format like Zarr.  You can read more about Zarr here ...
13.  We'll pick up from here next time!

mrocklin (commented):
Thanks for pushing up this draft @jakirkham !  

As usual, I have a ton of tiny comments, a few questions, and a couple larger comments.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
jakirkham: Sure I'll give this another look to fix text like this. Thanks!

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
jakirkham: Currently, yes. We could certainly change that. Though I regularly encountered use cases before when the image files did not necessarily have the same size.

Perhaps we could say, "if we know the image files to be uniform, we could inspect one for metadata and apply this to the rest like so."

Should add this is one of the things that should be in future work. In particular having a fast way to determine the metadata without reading the full image file into memory. Maybe we should hint at that here. Under the hood dask-image's `imread` does this and its quite fast, but doesn't handle all image file types.

xref: https://github.com/imageio/imageio/issues/382
xref: https://github.com/imageio/imageio/issues/362
xref: https://github.com/imageio/imageio/issues/263

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
jakirkham: Yep, we could do that. Should we keep this code or merely reference it in text?

jakirkham:
Thanks for the great feedback, Matt! 😄 Have committed all the small things and followed up on a couple of threads. Will take a look at some of the bigger items next.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
mrocklin: Any thoughts on this layout @jakirkham ?

guillaumeeb:
Hi,

What about some opening or even a section on Xarray and rasterio? This may be satellite imagery specific, but this is something we really need to do well from a Pangeo/CNES perspective.

cc @jmichel-otb, but this may be out of scope of this post.

jakirkham:
Thanks for the suggestion. I think I won't dive too deeply into domain specific tooling here. Trying to keep this relatively general. Though wouldn't be opposed to linking out to a blogpost that delved into more detail for the satellite imagery field. Are you interested in writing such a blog, @guillaumeeb?

guillaumeeb:
I'm not expert enough in the field, but @scottyhq [did it last year](https://medium.com/pangeo/cloud-native-geoprocessing-of-earth-observation-satellite-data-with-pangeo-997692d91ca2), maybe you can point to that?

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
jakirkham: Have added some description of these parameters above. Feel free to take a look. Though I probably wouldn't get too focused on details here as this will almost certainly vary across datasets.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
jakirkham: Any follow-up here, @mrocklin?

mrocklin: > Perhaps we could say, "if we know the image files to be uniform, we could inspect one for metadata and apply this to the rest like so."

Sounds good to me.

> Under the hood dask-image's imread does this and its quite fast, but doesn't handle all image file types.

Ah, I assumed thhat it would be reading all of the data.  If it's fast, then great.  

I still think that I have a slight preference for delayed just because it's a bit more generalizable and explicit, but whatever works.

> 
```suggestion
    data.append(dask.array.imread(fn))
```

jakirkham: Ok, I'll give that a go.

Yeah dask-image's reader is fast, but `dask.array`s is not. We could change that in the example. I think I noted this in "Future Work" currently, but we could replace that.

Sure one thought on the progression might be...

1. Load file with `skimage.io.imread`.
2. Note this won't scale to all of our data.
3. Show example with `delayed`.
4. We can brush this under the rug with dask-image's `imread`.

Does that seem reasonable?

Side note: The suggested change doesn't seem to have anything in the diff?

mrocklin: Yeah, that sounds great.   We're explicit so readers can generalize, but also point to the existence  of an automatic solution so that users know where to investigate further.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
jakirkham: Thanks for the feedback here.

1. I'd like to keep the series overview at least in this post as it will be the first one.
2. It probably has a fair bit of room for cutting (I have a tendency to be a bit verbose).
3. Happy to break this up with more code, pictures, diagrams, etc.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
jakirkham: I've added some code before this showing delayed. Have also reworked this section to match [this outline]( https://github.com/dask/dask-blog/pull/30#discussion_r292674949 ).

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
jakirkham: Ok, I've actually done this above with loading a single image and just changed the code here to use `imread`.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
jakirkham: Thanks for writing this up.

Sorry for being slow. Yeah I have been thinking about this and am still thinking about it.

It's an interesting approach and I can certainly see the appeal. From a high level it seems generally reasonable. Some of the details we may still need to sort out.

I'm still trying to figure out how we can leverage as much of the existing text as possible here while moving towards this structure. So have been working through some of the smaller comments that move in this direction.

Will try to restructure things a bit more based on this feedback and then we can see where things stand.

jakirkham: Also on the `concatenate`, `stack`, `block` progression, I've added some text and code along these lines. Could use a look when you have a chance.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
```diff
@@ -0,0 +1,226 @@
...
+scientists to leverage different tools to move towards a high level, friendly,
+cohesive, interactive analytical pipeline.
+
+Executive Summary
+-----------------
+
+This post explores a few simple workflows that a user might leverage to load
+their image with Dask. At the end of this post users should be able to
+construct a Dask Array, which they can use to further explore and manipulate
+their image data.
```
mrocklin: I recommend placing this at the beginning of the post, before the series overview.

Many blogpost visits are less than 10s long.  You need to hook them in those 10s.  If you don't then they leave.  I recommend having this summary, plus an exciting image on the screen when they first load the page.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
mrocklin: > I'd like to keep the series overview at least in this post as it will be the first one.

My experience with this is that I only finish about half of the blogpost series that I intend to write.  Other things get in the way.  I understand leaving this in if you're going to add more posts with high probability within a week or two.  Our timeline doesn't seem to be this fast or this predictable though, so I recommend leaving it off until we actually have a series.  Even then, I would reduce the series summary to a few sentences followed by a bulleted list of posts in the series.

jakirkham:
@mrocklin, could you please give this another look and share some thoughts on how it is going? The larger comment that you had above is at best partially addressed at this point, but many of the other comments should have been addressed. Thanks in advance 🙂

mrocklin:
> could you please give this another look and share some thoughts on how it is going? The larger comment that you had above is at best partially addressed at this point, but many of the other comments should have been addressed. Thanks in advance

I think it looks pretty clean.  I do think that we could make it flashier by adding a few things around Gokul's image data.  Any objection if I take a crack at adding a few things in a couple days?

jakirkham:
Thanks for giving it another look. As noted that part hasn't been addressed, I'm happy to work on that.

I'd prefer to work on it myself. Would like the experience and the satisfaction of having completed this.

mrocklin (commented):
Two small notes on starting things out.

For next steps, if you agree with the plan in the large comment, I would recommend:

1.  Adding a couple of the IndeX images to the front. 
2. Adding a couple of sentences about the data, along with a scary-but-truncated `ls` output
3.  Replacing the filenames `"mydata_ch<i>_<j>t_<k>x_<l>y_<m>z.tif "` with the actual filenames from the Gokul dataset
4. Keep with the concatenate/stack examples to start (even if they're not accurate) but include the output of `_repr_html_` as HTML text within this post
5. Include the full example with `da.block`, but perhaps don't talk about it much, saying instead "in reality things were more complex, so we arranged our delayed reads into a numpy array, and then used the [da.block]() function (providing actual link).
6.  Show full dask array image using `_repr_html_`

My hope is that this doesn't have to take too much time, both because there doesn't have to be a ton of prose around this, and because you have most of the code.

_posts/2019-06-04-loading-large-image-data-with-dask-array.md
mrocklin: If we have uniform shapes and dtypes then I think that there is a more effective pattern than what is shown above. 

```python
sample = skimage.io.imread(filenames[0])
```

Show output of single sample image


One sentence about how we can now lazily read in all of the files as lazy single-block dask arrays, using the sample as a pattern.

```python
arrays = [da.from_delayed(dask.delayed(skimage.io.imread)(fn), shape=sample.shape, dtype=sample.dtype) for fn in filenames]
```

One or two sentences about how we can then concatenate or stack these arrays into a larger dask array, followed by tiny code examples with da.concatenate and da.stack and their visualized shapes

jakirkham:
Have pinged a few people to take a look and share feedback. Are there any other corrections we should handle? When would we like to publish this?

_posts/2019-06-19-load-image-data.md
```diff
@@ -0,0 +1,1095 @@
...
+theme: twitter
+---
+{% include JB/setup %}
+
+Executive Summary
+-----------------
+
+This post explores simple workflows to load large stacks of image data with Dask array.
+
+In particular, we start with a directory full of TIFF files of images like the
```
rxist525: perhaps link a folder with sample data? The folks following along can download the exact set of files from here
https://drive.google.com/open?id=12UE8ZNmcnDFxnhvx1WRc0tT3yBrsRjgr

mrocklin: Whoa, Gokul joined GitHub :)  Welcome!

_posts/2019-06-19-load-image-data.md
```diff
@@ -0,0 +1,1095 @@
...
+ex6-2_CamA_ch1_CAM1_stack0003_560nm_0131245msec_0001423040msecAbs_000x_000y_000z_0000t.tif
+ex6-2_CamA_ch1_CAM1_stack0004_560nm_0174993msec_0001466788msecAbs_000x_000y_000z_0000t.tif
+```
+
+### Load a single sample image with Scikit-Image
+
+To load a single image, we use [Scikit-Image](https://scikit-image.org/):
+
+```python
+>>> import glob
```
rxist525: import os
os.chdir("path")

_posts/2019-06-19-load-image-data.md
```diff
@@ -0,0 +1,1095 @@
...
+
+
+on which some day we'll eventually be able to perform complex calculations.
+
+<img src="https://raw.githubusercontent.com/mrocklin/raw-host/gh-pages/images/aollsm-index-1.jpg"
+     width="45%"
+     alt="Light Microscopy data rendered with NVidia IndeX">
+<img src="https://raw.githubusercontent.com/mrocklin/raw-host/gh-pages/images/aollsm-index-2.jpg"
+     width="45%"
+     alt="Light Microscopy data rendered with NVidia IndeX">
```
thewtex: Text that describes what kind of data it is, e.g. light microscopy of ..., and what we would like to learn from it could provide context and motivation for a reader.

rxist525: Imaging the organelles of precursor neuronal cells in the hind brain of a developing zebrafish embryo using Lattice light sheet microscopy data with adaptive optics. 
Data source: 
Figure 3; Movie 4
https://science.sciencemag.org/content/360/6386/eaaq1392
pre-print for folks that can't access science journal
https://www.biorxiv.org/content/10.1101/243352v2

thewtex: @rxist525 awesome!

_posts/2019-06-19-load-image-data.md
```diff
@@ -0,0 +1,1095 @@
...
+Saving Data
+-----------
+
+To simplify data loading in the future, we store this in a large chunked
+array format like [Zarr]( https://zarr.readthedocs.io/ ) using the [to_zarr](
+https://docs.dask.org/en/latest/array-api.html#dask.array.Array.to_zarr )
+method.
+
+```python
+a.to_zarr("mydata.zarr")
```
thewtex: @jakirkham I am sure you have already thought of this, but it would be wonderful to preserve the [image metadata](https://github.com/imageio/imageio/issues/382) in attributes when saving to zarr, i.e. the spatial origin and spacing.

_posts/2019-06-19-load-image-data.md
mrocklin: I've added links to the paper and preprint, but also added a disclaimer that this workload should be generally applicable.

_posts/2019-06-19-load-image-data.md
mrocklin: I've changed the glob command from `glob("raw/*.tif")` to `glob("/path/to/raw/*.tif")` to hopefully avoid the confusion that I think you're concerned about.

_posts/2019-06-19-load-image-data.md
mrocklin: I've gone and added this link in a couple places.  Thanks @rxist525

mrocklin:
I think that we should merge and publish.  @jakirkham I'll wait until you wake up and give this the final OK.

_posts/2019-06-19-load-image-data.md
jakirkham: Thanks @thewtex. Added a small note about leveraging attributes to store metadata. 🙂

jakirkham:
Thanks all. Also thanks Matt for fielding the last round of edits. 😄

LGTM.

_posts/2019-06-20-load-image-data.md
```diff
@@ -0,0 +1,1115 @@
...
+on which some day we'll eventually be able to perform complex calculations.
+
+<img src="https://raw.githubusercontent.com/mrocklin/raw-host/gh-pages/images/aollsm-index-1.jpg"
+     width="45%"
+     alt="Light Microscopy data rendered with NVidia IndeX">
+<img src="https://raw.githubusercontent.com/mrocklin/raw-host/gh-pages/images/aollsm-index-2.jpg"
+     width="45%"
+     alt="Light Microscopy data rendered with NVidia IndeX">
+
+*Disclaimer: we're not going to produces rendered images like the above in this
```
jakirkham, line 114: Changing "produces" to "produce". ( https://github.com/dask/dask-blog/pull/36 )

PR #31, opened 2019-06-17
Use pkg-config (instead of pkgconfig)

jakirkham:
The former is a Unix build tool, which is needed when building things that leverage autotools. The latter is a Python package for interacting with that Unix build tool. We are only using the former in our UCX build.

ref: https://github.com/conda-forge/pkgconfig-feedstock/issues/4

### DIFF
--- a/_posts/2019-06-09-ucx-dgx.md
+++ b/_posts/2019-06-09-ucx-dgx.md
@@ -511,7 +511,7 @@ Create Conda Environment
 ------------------------
 
 ```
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @jakirkham

jakirkham:
On a different note, I'm wondering if we should drop the `--enable-debug` flag from `../configure` below. This is mainly needed by developers, but not necessarily others. Thoughts?

mrocklin:
I don't have any thoughts here.

I think that the long-term value of these instructions are low, so I
wouldn't invest too much into them.

PR #37, opened 2019-07-04
Add custom CuPy kernels with Dask post

pentschev:


pentschev:
cc @mrocklin

mrocklin:
Nice post!  I like that it's quick and focused.  The examples seem clear to me and easy to follow from the perspective of someone who doesn't know much about CUDA.

Two minor comments/requests from a Dask perspective:

1.  I have a slight preference for us to avoid `da.from_array` and instead prefer to use `da.random.random`.  My reasoning here is that from repeatedly reading examples with `da.from_array`, users might decide that it's a good idea for them to do on their own, and that's probably not a good idea in most use cases in terms of performance or managing memory.
2.  I'm slightly concerned that we're mutating blocks within a dask array.  Dask tasks are supposed to not have side effects like these.  Perhaps we can allocate a cupy array within the supporting function?

pentschev:
> 1\. I have a slight preference for us to avoid `da.from_array` and instead prefer to use `da.random.random`.  My reasoning here is that from repeatedly reading examples with `da.from_array`, users might decide that it's a good idea for them to do on their own, and that's probably not a good idea in most use cases in terms of performance or managing memory.

I would prefer a deterministic computation rather than a random array for two reasons:

1. It's clearer to show that the result is correct without having to reverse the operation;
1. Using only `da.random.random` makes it unclear how to create Dask+CuPy arrays from real data, since random arrays have a very limited scope.

Do we have another way of creating non-random Dask+CuPy arrays? I would be totally up for changing it to something like that.

> 2\. I'm slightly concerned that we're mutating blocks within a dask array.  Dask tasks are supposed to not have side effects like these.  Perhaps we can allocate a cupy array within the supporting function?

Sure, this can be changed easily.

mrocklin:
> I would prefer a deterministic computation rather than a random array for two reasons:

Agreed on your first point.  I can see why it would be nice to have the same data both with and without Dask.  You could use `da.arange(...).reshape(...)` as well, though I imagine that the reshaping might be troublesome.  I'm also fine if you ignore the request (perhaps it doesn't make sense to worry about too much).

For your second point:

> makes it unclear how to create Dask+CuPy arrays from real data, since random arrays have a very limited scope

This is also my concern about `da.from_array` with in-memory data.  This isn't really a way that we would want people to use Dask.  There are quite a few issues and stack overflow questions out there saying things like "I made this giant numpy array/pandas dataframe and gave it to Dask but I still ran out of memory."  It's because of these user issues that I'd like to diminish the use of `dd.from_pandas` and `da.from_array` where convenient.  Perhaps in this case though it's not convenient, which is fine.

pentschev:
> You could use da.arange(...).reshape(...) as well, though I imagine that the reshaping might be troublesome.

But can we use `da.arange(...)` to create a CuPy array?

> This is also my concern about `da.from_array` with in-memory data. This isn't really a way that we would want people to use Dask. There are quite a few issues and stack overflow questions out there saying things like "I made this giant numpy array/pandas dataframe and gave it to Dask but I still ran out of memory." It's because of these user issues that I'd like to diminish the use of `dd.from_pandas` and `da.from_array` where convenient. Perhaps in this case though it's not convenient, which is fine.

I understand your point. In that case, I would say we should work on permitting Dask to create arrays with different backends. A simple solution (though NumPy-incompatible API) would be to add a `like=cupy.array` argument, or perhaps `meta=cupy.array`? How do you feel about adding a convenience such as that to Dask?

mrocklin:
Something like that seems reasonable.  It's probably not a super high priority though.  I say that we just use your original solution of `from_array` for now.

mrocklin:
Or I should say, it doesn't seem that high priority to me, but I might be missing things.

pentschev:
I think it makes sense to have something like that, but as of now we don't really have a real alternative to `from_array` for cases like this. I would say here we have two alternatives to make things clearer in the post:

1. Add a note stating that `from_array` is not a nice solution long-term and say we will work on improving that; or
2. Say nothing now and work on creating arrays with Dask but using different backends, and then publish a post about that new feature.

Priority is harder to define, it really depends on how many people need/want to use Dask with CuPy as a backend at the moment.

mrocklin:
I say that we just drop my original comment and move on to other things.

> depends on how many people need/want to use Dask with CuPy as a backend at the moment.

Well even then I would expect them to read data from some data source.  I'm not sure that arange gets used for much outside of examples.

pentschev:
> Well even then I would expect them to read data from some data source. I'm not sure that arange gets used for much outside of examples.

Yes, that's true. I was thinking of array creation functions in general, but there aren't too many anyway.

pentschev:
I updated the post now to create the output array inside `dispatch_add_kernel`, and as per our discussion, left `from_array` as is.

_posts/2019-07-08-dask-custom-cupy-kernel.md
```diff
@@ -0,0 +1,208 @@
...
+tags: [Dask, CuPy]
+theme: twitter
+---
+{% include JB/setup %}
+
+
+Summary
+-------
+
+About three months ago, [Matthew Rocklin](https://twitter.com/mrocklin) wrote a
```
mrocklin:
```suggestion
About three months ago, we wrote a
```

pentschev: You always give nominal credit to everybody, why don't you want that for yourself?

pentschev: I can remove the link or change it to github profile if you prefer too.

pentschev:
This has been published in medium.com: https://medium.com/rapids-ai/parallelizing-custom-cupy-kernels-with-dask-4d2ccd3b0732 . Closing.

PR #38, opened 2019-07-05
[WIP] Add large svd draft

mrocklin:
cc @alimanfoo

I need to put in actual code samples and probably do a screencast, but in the mean time if you'd be willing to look through this post for the general structure I'd welcome the collaboration.

### DIFF
--- a/_posts/2019-07-02-very-large-svds.md
+++ b/_posts/2019-07-02-very-large-svds.md
@@ -121,7 +121,7 @@ Dask array has one of these approximation algorithms implemented in the
 function.  And with it we can compute the approximate SVD of very large
 matrices.
 
-We were recently working on a problem (explained below) and found that I was
+We were recently working on a problem (explained below) and found that we were
 still running out of memory when dealing with this algorithm.  There were two
 challenges that we ran into:
 
@@ -241,18 +241,22 @@ It takes around 2min 30s time to compute that on a laptop.  That's great!  But w
 Adding GPUs (a 15 second SVD)
 -----------------------------
 
-We can increase performance considerably by using a multi-GPU machine.
-Here is almost the same code, running in significantly less same time but we make the
+*Disclaimer: one of the authors (Ben Zaitlen) works for NVIDIA*
+
+We can dramatically increase performance by using a multi-GPU machine.  NVIDIA and other manufactures now make machines with multiple GPUs co-located in the same physical box.  In the following section, we will run the calculations on a **DGX2**, a machine with 16 GPUs.
+
+
+Below is almost the same code, running in significantly less same time but we make the
 following changes:
 
 1.  We increase the size of the array by a factor of **10x**
 2.  We switch out NumPy for CuPy, a GPU NumPy implementation
-3.  We use a sixteen-GPU DGX-2 machine with NVLink interconnects between GPUs (this will dramatically decrease transfer time between workers)
+3.  We use a sixteen-GPU DGX-2 machine with NVLink interconnects between GPUs (NVLink will dramatically decrease transfer time between workers)
 
 On A DGX2 we can calculate an SVD on a 200GB Dask array between 10 and15 seconds: [SVD Multi-GPU Notebook](https://gist.github.com/quasiben/98ee254920837313946f621e103d41f4)
 
 To see this run, we recommend viewing
-[the attached screencast](https://youtu.be/4X5yky2lvEw)
+[the attached screencast](https://www.youtube.com/watch?v=6hmt1gARqp0)
 
 
 Read dataset from Disk
@@ -275,9 +279,8 @@ The combination of preprocessing and SVD calculations ran in 18.7 sec and the da
 
 Again, on a DGX2, from data loading to SVD we are running in time less than it would take to make a cup of tea.
 However, the data loading can be accelerated.
-From GCS we are reading into host memory, uncompressing the zarr bits,
-then moving the data from host memory to device memory.
-This is be improved if we cut out the host entirely and read directly into the GPU.
+From GCS we are reading into data into the main memory of the machine (host memory), uncompressing the zarr bits,
+then moving the data from host memory to the GPU (device memory).  Passing data back and forth between host and device memory can significantly decrease performance.  Reading directly into the GPU, bypassing host memory, would improve the overall pipeline.
 
 And so we come back to a common lesson of high performance computing:
 
--- a/_posts/2019-07-02-very-large-svds.md
+++ b/_posts/2019-07-02-very-large-svds.md
@@ -2,7 +2,7 @@
 layout: post
 title: Very Large SVDs
 tagline: Dask + CuPy + Zarr + Genomics
-author: Matthew Rocklin
+author: Matthew Rocklin & Alistair Miles
 tags: [GPU, array, cupy]
 draft: true
 theme: twitter
@@ -145,8 +145,8 @@ Application - Genomics
 ----------------------
 
 Many studies are using genome sequencing to study genetic variation
-one from each parent.  For those not biologist, recall the [Punnett square](https://en.wikipedia.org/wiki/Punnett_square), where	between different individuals within a species.  These includes
-each box is filled letter from each column/row combination.	studies of human populations, but also other species such as mice,
+between different individuals within a species.  These includes
+studies of human populations, but also other species such as mice,
 mosquitoes or disease-causing parasites.  These studies will, in
 general, find a large number of sites in the genome sequence where
 individuals differ from each other.  For example, humans have more
@@ -158,25 +158,23 @@ In diploid species like humans, mice or mosquitoes, each individual
 carries two genome sequences, one inherited from each parent.  At each
 of the 100 million variable genome sites there will be two or more
 "alleles" that a single genome might carry.  One way to think about
-this is via the [Punnett square](https://en.wikipedia.org/wiki/Punnett_square), which
+this is via the [Punnett
+square](https://en.wikipedia.org/wiki/Punnett_square), which
 represents the different possible genotypes that one individual might
-carry at one of these variable sites
+carry at one of these variable sites:
+
 <td>
 <img src="https://upload.wikimedia.org/wikipedia/commons/9/93/Punnett_Square_Genetic_Carriers.PNG" alt="punnet square" height="40%" width="40%">
 </td>
 
 In the above there are three possible genotypes: AA, Aa, and aa.  For
-
 computational genomics, these genotypes can be encoded as 0, 1, or 2.
-Humans have 20,000-25,000 genes and experiments across populations can vary between 10s of thousand of individuals to 10 Million individuals.  For example,  [biobank](https://www.ukbiobank.ac.uk/) SOMETHING ABOUT THE SCIENCE	In a study of a species with M genetic variants assayed in N
-
+In a study of a species with M genetic variants assayed in N
 individual samples, we can represent these genotypes as an (N x M)
-Common questions we want to answer of this data include: How do genes vary across populations ? Can we classify a set of genes for a particular trait ?  [Performing an SVD](https://www.pnas.org/content/97/18/10101) helps scientists understand what is happening, not at an individual gene level, but across the entire genome.	array of integers.  For a modern human genetics study, the scale of
-
+array of integers.  For a modern human genetics study, the scale of
 this array might approach (100 million x 1 million).  (Although in
-Reducing the time computing, like all science, allows for testing more hypotheses in less time.	practice, the size of the first dimension (number of variants) can be
-Practically, this means not simply a fast SVD but an accelerated pipeline end-to-end, from data loading to analysis, to understanding.	reduced somewhat, by at least an order of magnitude, because many
-
+practice, the size of the first dimension (number of variants) can be
+reduced somewhat, by at least an order of magnitude, because many
 genetic variants will carry little information and/or be correlated
 with each other.)
 
@@ -184,7 +182,8 @@ These genetic differences are not random, but carry information about
 patterns of genetic similarity and shared ancestry between
 individuals, because of the way they have been inherited through many
 generations.  A common task is to perform a dimensionality reduction
-analysis on these data, such as a [principal components analysis](https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.0020190)
+analysis on these data, such as a [principal components
+analysis](https://journals.plos.org/plosgenetics/article?id=10.1371/journal.pgen.0020190)
 (SVD), to identify genetic structure reflecting these differencies in
 degree of shared ancestry.  This is an essential part of discovering
 genetic variants associated with different diseases, and for learning
--- a/_posts/2019-07-02-very-large-svds.md
+++ b/_posts/2019-07-02-very-large-svds.md
@@ -170,7 +170,7 @@ carry at one of these variable sites:
 In the above there are three possible genotypes: AA, Aa, and aa.  For
 computational genomics, these genotypes can be encoded as 0, 1, or 2.
 In a study of a species with M genetic variants assayed in N
-individual samples, we can represent these genotypes as an (N x M)
+individual samples, we can represent these genotypes as an (M x N)
 array of integers.  For a modern human genetics study, the scale of
 this array might approach (100 million x 1 million).  (Although in
 practice, the size of the first dimension (number of variants) can be
--- a/_posts/2019-07-02-very-large-svds.md
+++ b/_posts/2019-07-02-very-large-svds.md
@@ -262,9 +262,9 @@ In practice though, our input array won't be randomly generated, it will be
 coming from some dataset stored on disk or increasingly more common, stored in the cloud.
 To make things more realistic we perform a similar calculation with data stored in a [Zarr format](https://zarr.readthedocs.io/en/stable/) (which is what Alistair, our genomics collaborator, uses) in [GCS](https://cloud.google.com/storage)
 
-In this [Zarr SVD example](https://gist.github.com/quasiben/e52bc740ae22ae321f30987c65998078), we load a 25GB GCS backed data set onto a DGX2,
+In this [Zarr SVD example](https://gist.github.com/quasiben/e52bc740ae22ae321f30987c65998078), we load a 25GB GCS backed data set onto a DGX2, run a few processing steps, then perform an SVD.  The combination of preprocessing and SVD calculations ran in 18.7 sec and the data loading took 14.3 seconds.
 
-Again, on a DGX2, from data loading to SVD we are running in time less than it take to make a cup of tea.
+Again, on a DGX2, from data loading to SVD we are running in time less than it would take to make a cup of tea.  However, the data loading can be accelerated.  From GCS we are reading into host memory, uncompressing the zarr bits, then moving the data from host memory to device memory.  This is be improved if we cut out the host entirely and read directly into the GPU.
 
 And so we come back to a common lesson of high performance computing:
 
--- a/_posts/2019-07-02-very-large-svds.md
+++ b/_posts/2019-07-02-very-large-svds.md
@@ -1,0 +1,253 @@
+---
+layout: post
+title: Very Large SVDs
+tagline: Dask + CuPy + Zarr + Genomics
+author: Matthew Rocklin
+tags: [GPU, array, cupy]
+draft: true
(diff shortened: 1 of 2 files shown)


alimanfoo:
Cool, thanks @mrocklin. I took a scan through and looks good, I'll try to follow up with more detailed feedback.

mrocklin:
If you want to give it a try, see this notebook https://gist.github.com/mrocklin/3af8a428e18a3047ab501161ffe17cf9

It should run fine on the single-node GPU machine to which you have access today (sending the address by e-mail)

pentschev:
No `%%time`? :(

mrocklin:
@quasiben @jakirkham @pentschev I suspect that you all are busy, but it would be fun to try this experiment again, now with the new and improved UCX integration.

quasiben:
I gave this a whirl on a DGX2 (16GPUs) with NVLink enabled. Here's a performance report:

https://gistcdn.githack.com/quasiben/78033ef8130afb7b4c5f5f2def36d388/raw/59eb22d47bef5462bdb433ebfe3ae099cc64ccf1/dask-performance-large-svd.html

In general things are looking ok -- I am a bit concerned with the NumPy transfers (don't know where they are coming from yet) and there is a fairly large gap in the middle of the workflow.  But things worked!

pentschev:
> I am a bit concerned with the NumPy transfers (don't know where they are coming from yet)

Probably metadata in UCX send/recv, which is mostly being addressed in https://github.com/dask/distributed/pull/3732 .

mrocklin:
> I am a bit concerned with the NumPy transfers (don't know where they are coming from yet)

It would be nice to know how much was transferred (this is mostly a mental note to me to maybe improve diagnostics)

mrocklin:
Can I ask that you maybe also call `v.compute()`.  That might slightly change things.

quasiben:
> Can I ask that you maybe also call v.compute(). That might slightly change things.

Do you mean `u.compute()`  ?  I called `v`

mrocklin:
From the performance report it looks like we're spending essentially all of our time generating random numbers.  

There are a few things that we might experiment with here to improve the focus on SVD, and also give this example a bit more reality.  cc'ing @alimanfoo, @eric-czech, and @jakirkham who I think might all care about this problem.

1.  We can mimic a genomics problem I think by using random integers uniformly distributed between 0 and 3, stored as uint8.
2.  We might consider storing these somewhere.  We could try something small that fits in GPU memory (it might be easier given that we're storing these somewhat compactly).  So we would persist the input dask array fill up memory (we probably have to reduce size a bit) and then and then call svd, hoping that the working memory that it needs wasn't too much greater.
3.  For something larger, we could move to disk.  I *suspect* that reading from disk will be more expensive than regenerating the data, but this is still a useful thing to highlight and will make a nice section.  `<puts on NVIIDA hat>` It's also a great opportunity to mention NVIDIA IO technologies `</takes off NVIDIA hat>`
4.  If we wanted to experiment a bit, we could play with compressing the data in memory just before we persist it.  This would require us to have functions that converted a cupy array to a compressed form, and then decompressed that cupy array.  Do we have such a function?
5.  We could fit everything into memory if we just added more machines :)

Anyway, those are some ideas.  I'm actually pretty excited about trying them all.

mrocklin:
@quasiben I'm somewhat concerned about the 2GB/s bandwidth for cupy arrays.  This seems low?

I hope that once we remove the random number generation that this becomes our next bottleneck to squash.

mrocklin:
Oh indeed you did.  My apologies.  v is good.  u is probably huge.

alimanfoo:
> 1. We can mimic a genomics problem I think by using random integers uniformly distributed between 0 and 3, stored as uint8.

FWIW in a real genomics problem the integers can take values 0, 1 or 2 (if working with a diploid organism like humans or mosquitoes). They are also not uniformly distributed. These may or may not make a different, depending on what you're testing. Biggest different I expect would be compressibility (real data will be much more compressible than uniform random). Give me a shout if you'd like some real data to play with, I can point you at some zarrs on GCS.

mrocklin:
Some zarr datasets on gcs would be great.  We could probably download
locally.

mrocklin:
Here is an example using Zarr with dask array together to compress in-memory data.

https://gist.github.com/mrocklin/ae18a1b51ccf90a2fa9872de23144ffa

alimanfoo:
Here's a gist showing location of some genome variation data in zarr on GCS, and also the couple of preprocessing steps required before can be run through SVD:

https://gist.github.com/alimanfoo/ef6d9e70750b5273f95cbf681d162de8

Hth :-)

alimanfoo:
> Here's a gist showing location of some genome variation data in zarr on GCS, and also the couple of preprocessing steps required before can be run through SVD:
> 
> https://gist.github.com/alimanfoo/ef6d9e70750b5273f95cbf681d162de8
> 
> Hth :-)

Btw this example is "tall and skinny" with ~10 million features and ~1,000 individuals, but the data are growing to become more square, e.g., biobank datasets are more like ~10 million features and ~1 million individuals. So don't optimise for the tall and skinny case. At some point would probably be useful to simulate a biobank-scale dataset.

alimanfoo:
> Here is an example using Zarr with dask array together to compress in-memory data.
> 
> https://gist.github.com/mrocklin/ae18a1b51ccf90a2fa9872de23144ffa

That's cunning :-)

mrocklin:
> That's cunning :-)

Yeah, I then followed it up by actually swapping out zarr with functions that do bit-twiddling

```python
def compress(x: np.ndarray) -> np.ndarray:
   out = np.zeros_like(x, shape=(x.shape[0], x.shape[1] // 4))
   out += x[:, 0::4]
   out += x[:, 1::4] << 2
   out += x[:, 2::4] << 4
   out += x[:, 3::4] << 6
   return out

def decompress(out: np.ndarray) -> np.ndarray:
   back = np.zeros_like(out, shape=(out.shape[0], out.shape[1] * 4))
   back[:, 0::4] = out & 0b00000011
   back[:, 1::4] = (out & 0b00001100) >> 2
   back[:, 2::4] = (out & 0b00110000) >> 4
   back[:, 3::4] = (out & 0b11000000) >> 6
   return back
```

This is a bit faster, and it also works wtih CuPy (we don't have good compression algorithms with Zarr on GPU yet).  This lets us store a dataset pretty compactly, and expand it on-demand.  It didn't quite work, I suspect because of a known memory pressure issue with the dask scheduler (a few other groups have run into something similar, so I hope that there is good pressure to solve it).  

Speaking of CuPy, @quasiben and I were playing with a DGX-2 (16GPUs, half terrabyte of device memory) last night and solved something like a 10,000,000 by 20,000 svd in 10 seconds or so (if I recall correctly).  That's after it was in RAM though.  Probably the main cost will be collecting it from wherever it's stored.

> Btw this example is "tall and skinny" with ~10 million features and ~1,000 individuals, but the data are growing to become more square, e.g., biobank datasets are more like ~10 million features and ~1 million individuals. So don't optimise for the tall and skinny case. At some point would probably be useful to simulate a biobank-scale dataset

That's good to know.

So far it looks like we're not overly sensitive to chunking on the y-axis, assuming that we're comfortable with approximate results.

eric-czech:
> swapping out zarr with functions that do bit-twiddling

@mrocklin do you have any thoughts on trying to do bitpacking like that with the dask API rather than as `map_blocks` functions (or zarr filters, even assuming they worked with cupy)?  Looking at that example, I'm immediately seeing some ways to do useful math on the packed representations so if applying the compression filters at that level instead and relying on bitwise da.Array functions rather than custom numpy functions isn't a bad idea, I'd like to explore that some more.

@alimanfoo mentioned experimenting with something like that in the past a bit but I'd love to know if you think doing lots of bit-fiddling at the dask level would be a bad practice.

Apologies for a somewhat off-topic post but your example shook some thoughts loose so thanks for sharing it!

mrocklin:
> do you have any thoughts on trying to do bitpacking like that with the dask API rather than as map_blocks functions

In principle, yes, this could be made to work.  Dask array isn't as smooth on in-place operations, so to make the packing process here as fast then we might want something like https://github.com/dask/dask/issues/6123 .  If you want to pack with map_blocks and then operate on the packed representation with the Dask array API though then yeah, I imagine that that would be easier.

mrocklin:
In general, the experiment with @quasiben was interesting in order to see how much of this data we could effectively operate on in a fixed amount of device memory.  Things are *very fast* if we can stay in that RAM, but there are a lot of interesting complications to doing so.  This was more for us to play with technology than to solve an actual science problem.  

This leads to a larger question of "what are the actual science problems here?"  For example, you all may not even care about doing this quickly, but instead just care about being able to do it at all.  I encourage folks to speak up about what would be motivating.

alimanfoo:
> This leads to a larger question of "what are the actual science problems here?" For example, you all may not even care about doing this quickly, but instead just care about being able to do it at all. I encourage folks to speak up about what would be motivating.

I'd say for genomics the main driver is to be able to do it at all. If it can then be done in an interactive environment, even better, where "interactive time" is anything less than the time it takes to make a cup of tea. I.e., you don't need to leave your normal working environment (Jupyter notebook) or worry about any long-running tasks. Also probably relevant as analysis moves to the cloud is cost, i.e., it is not an expensive computation. 

Sorry this is a bit vague, but hopefully useful. Main point is we are not running SVDs over and over again all day, so this is not something where there is strong pressure to run as fast and cheaply as possible. But making it possible and straightforward to setup and run large SVDs is valuable.

quasiben:
@alimanfoo can you make tea in 20 seconds ? 

In my last experiment I was able to push to 400GB of in memory data (10_000_000, 30_000).  SVD on this size was also not greatly impacted.  I measured a compute time of 19.3 seconds.  This is probably close to the limit (as of now) of the amount of data we can naively perform an SVD on.

To my snarky tea comment before, and as Matt said earlier, reading from disk will be somewhat burdensome.  It took roughly a minute to generate the data in memory -- this will mostly like be longer when reading from disk and push us into tea construction time.

There are two things which _could_ be done here to alleviate performance and memory constraints:
- [Optimized GPU readers from disk](https://devblogs.nvidia.com/gpudirect-storage/) which integrate seamlessly with Zarr or other formats
- Increase single GPU RAM (don't know how hard this is) or move to a multinode setup (I am working on this now)

Another interesting thing to note is that after loading data into memory, SVD is a communication heavy process.  I didn't expect that.  A distributed SVD spends more time transferring than computing.  Using a box like a DGX2 is hugely impactful here as we can leverage NVLink for communication between GPUs (NVLink has a 50GB/s bandwidth)

Breakdown of actions for an SVD
```
defaultdict(int,
            {'compute': 55.52175450325012,
             'transfer': 141.12734627723694,
             'disk-read': 0.00934290885925293})
```

I also wanted to note that working on these problems is also of general interest to NVIDIA.
 ClaraGenomics would be an example of NVIDIA in the genetics space: https://github.com/clara-genomics/Compute4COVID

mrocklin:
It would be fun, I think, to do this on a practical problem (like the zarr
dataset that @alimanfoo pointed to earlier) on both a consumer laptop and a
DGX-2, and quantify all of the costs involved (reading from disk, moving to
GPU, transfer time, compute time).  That might give us a sense about what a
range of performance looks like, and where technology efforts would need to
focus in the future to improve things.

alimanfoo:
> @alimanfoo can you make tea in 20sec ?

Nope :-) Just to elaborate a little, obviously interactive is ideal, but tea break would be OK too.

Another way to think about it might be, imagine a researcher who has access to a pangeo-like or saturncloud-like research computing environment, and can fire up notebook servers and dask-kubernetes clusters as needed based on standard node types offered by public cloud providers. They need to run an SVD over a large dataset, and ideally it's (a) convenient to fire up the necessary compute resources and setup the computation, (2) doesn't take too long to run, and (iii) cloud costs are small enough to be negligible, i.e., don't require any special accounting. 

Hth.

mrocklin:
> Another way to think about it might be, imagine a researcher who has access to a pangeo-like or saturncloud-like research computing environment, and can fire up notebook servers and dask-kubernetes clusters as needed based on standard node types offered by public cloud providers. They need to run an SVD over a large dataset, and ideally it's (a) convenient to fire up the necessary compute resources and setup the computation, (2) doesn't take too long to run, and (iii) cloud costs are small enough to be negligible, i.e., don't require any special accounting.

Like this?

Notebook: https://nbviewer.jupyter.org/urls/gist.githubusercontent.com/mrocklin/0c4e97b9164c55d85ce457fce0b87d41/raw/90a07e2f6fd1563aa65eb83f190e4c853494a4dd/coiled-compressed-svd.ipynb
Screencast: https://www.youtube.com/watch?v=qaJcAvhgLy4

quasiben:
> Here's a gist showing location of some genome variation data in zarr on GCS, and also the couple of preprocessing steps required before can be run through SVD:

Starting from @alimanfoo 's notebook (above) I retooled a few things and re-ran on a DGX2:

- [notebook](https://gist.github.com/quasiben/3d03bc2b116162a43c9595a1de6a75e4)
- [performance report](https://gistcdn.githack.com/quasiben/58b97b691fc3cc603cb7d1e15bef1ad2/raw/c90a2453b913067a3ba365ae53739cf1faf3ff67/dask-zarr-svd-report.html)

The SVD took ~15-20 secs and data loading was in the 9-10sec range.  So we are still well within tea/interactive time

I also filed a dask issue related to the compress/slicing required. 
https://github.com/dask/dask/issues/6169

For now, I worked around it by casting to numpy

alimanfoo:
Thanks @quasiben, very interesting. Nice to see that data loading is not taking too long.

alimanfoo:
Btw I guess if data loading isn't too slow then calling persist() isn't strictly necessary? In which case this could scale well to larger datasets (too big to hold in memory)?

quasiben:
@alimanfoo that's correct.  The persist call was mostly to separate timings for data loading and cleaning/SVD.  How large are your datasets ?

alimanfoo:
> @alimanfoo that's correct. The persist call was mostly to separate timings for data loading and cleaning/SVD. How large are your datasets ?

For mosquitoes, we're up to order of 10,000,000 genetic variants (features) by 10,000 individual mosquitoes (observations). For human genetics, folks like @eric-czech are probably wanting to analyse datasets of the order of 10,000,000 genetic variants by 1,000,000 individual people ("biobank-scale"). Hth.

quasiben:
I don't have write access to this repo so I can't commit directly but I've started updates here:
https://github.com/quasiben/dask-blog/tree/svd

I'll push on this more later this afternoon.  @alimanfoo if you have time can you review the Genomics section ?

mrocklin:
You now have access to my fork

quasiben:
Thanks!  Pushed the commit which should be updated here as well

alimanfoo:
Added https://github.com/quasiben/dask-blog/pull/1 with a bit more explanation of the genomics.

quasiben:
Thanks @alimanfoo !  I added them directly here.  Matt gave me write access and thought it best to centralize now that we can

_posts/2019-07-02-very-large-svds.md
```diff
@@ -0,0 +1,271 @@
...
+carry at one of these variable sites:
+
+<td>
+<img src="https://upload.wikimedia.org/wikipedia/commons/9/93/Punnett_Square_Genetic_Carriers.PNG" alt="punnet square" height="40%" width="40%">
+</td>
+
+In the above there are three possible genotypes: AA, Aa, and aa.  For
+computational genomics, these genotypes can be encoded as 0, 1, or 2.
+In a study of a species with M genetic variants assayed in N
+individual samples, we can represent these genotypes as an (N x M)
```
alimanfoo:
```suggestion
individual samples, we can represent these genotypes as an (M x N)
```

Sorry, got my dimensions the wrong way round.

quasiben:
@mrocklin when you have some time can you take a pass over this ?

_posts/2019-07-02-very-large-svds.md
```diff
@@ -0,0 +1,276 @@
...
+following changes:
+
+1.  We increase the size of the array by a factor of **10x**
+2.  We switch out NumPy for CuPy, a GPU NumPy implementation
+3.  We use a sixteen-GPU DGX-2 machine with NVLink interconnects between GPUs (this will dramatically decrease transfer time between workers)
+
+On A DGX2 we can calculate an SVD on a 200GB Dask array between 10 and15 seconds: [SVD Multi-GPU Notebook](https://gist.github.com/quasiben/98ee254920837313946f621e103d41f4)
+
+To see this run, we recommend viewing
+[the attached screencast](https://youtu.be/4X5yky2lvEw)
```
mrocklin: My computer may be borked?  Can someone else check if this has audio?

mrocklin:
Thanks @quasiben I took a look.  Some high level thoughts:

1.  Can you check the screencast and verify that it has audio.  It didn't play on my machine, but my audio has also been strange recently.  It didn't play on my phone though too, which has me concerned
2.  The GPU part of this post assumes knowledge of terms like DGX2 and host memory that we might not be able to rely on with the majority of our readership.  It might also be good to somehow bracket that section in "now let's enter GPU land (oh and as a disclaimer, one of the authors works for NVIDIA)"
3.  I'd like to also include the bits on compression, why it's useful and why it didn't work
4.  I'd also like to include bits on moving to the cloud (with appropriate disclaimer there as well)
5.  We should probably shift language from "I" to "we".  I did this in a few places as I was going through but I probably missed some.  I also added you to the author list, and decided to sort alphabetically by first name.

quasiben:
@mrocklin thanks for the review.  I just recently upgraded to Catalina and I think that changed various app permissions.  I'll re-record and upload.

I can also update the post with your suggestions as well.  Thanks

_posts/2019-07-02-very-large-svds.md
```diff
@@ -0,0 +1,292 @@
+---
+layout: post
+title: Very Large SVDs
+tagline: Dask + CuPy + Zarr + Genomics
+author: Alistair Miles (Oxford Big Data Institute), Ben Zaitlen (NVIDIA), Matthew Rocklin (Coiled)
```
alimanfoo:
```suggestion
author: Ben Zaitlen (NVIDIA), Matthew Rocklin (Coiled), Alistair Miles (Oxford University)
```

I don't deserve to be first author here :)

quasiben: I think we should do it alphabetically by first name.  I believe that happens to coincide with the current ordering of names

alimanfoo: That would feel a bit strange for me, my contribution to this nice work is very minor.

quasiben: changed the author order

_posts/2019-07-02-very-large-svds.md
```diff
@@ -0,0 +1,495 @@
...
+dask.config.set({"optimization.fuse.ave-width": 5})
+```
+
+Then things work fine
+---------------------
+
+We're going to try this SVD on a few different choices of hardware including:
+
+1.  A MacBook Pro
+2.  A DGX-2, an NVIDIA worksation with 16 high-end GPUs and fast network
```
quasiben:
```suggestion
2.  A DGX-2, an NVIDIA worksation with 16 high-end GPUs and fast interconnect
```

_posts/2019-07-02-very-large-svds.md
```diff
@@ -0,0 +1,495 @@
...
+Now that things work, we can experiment with different hardware.
+
+
+### Adding GPUs (a 15 second SVD)
+
+*Disclaimer: one of the authors (Ben Zaitlen) works for NVIDIA*
+
+We can dramatically increase performance by using a multi-GPU machine.
+NVIDIA and other manufactures now make machines with multiple GPUs co-located in the same physical box.
+In the following section, we will run the calculations on a **DGX2**, a machine with 16 GPUs and fast network connection.
```
quasiben:
```suggestion
In the following section, we will run the calculations on a **DGX2**, a machine with 16 GPUs and fast interconnect between the GPUs.
```

quasiben:
This is looking really good.  @mrocklin left a few small comments on the latest commits

quasiben:
Read through the post again and I'm good with posting.  @alimanfoo  ?

alimanfoo:
LGTM :-)

mrocklin:
I will plan to merge this and tweet about it tomorrow morning US Eastern time

PR #45, opened 2019-07-31
2019 survey results

TomAugspurger:
This post is autogenerated from https://github.com/TomAugspurger/dask-survey-results/blob/master/analyze.ipynb.

I'm fighting with 

1. The CSS of the tables
2. The SVG plots

But comments on the content would be welcome.


TomAugspurger:
Fixed the formatting and the plotting issues.

https://mybinder.org/v2/gh/TomAugspurger/dask-survey-results/master?filepath=analyze.ipynb has a binder link. This post is essentially that, minus the code inputs.

mrocklin:
The binder is super fun!  I enjoyed looking at the "If you use a cluster,
how do you launch Dask" question with the background gradient colored by
rows rather than columns.  It was nice to be able to easily swap that in
and out.

It was also fun to use that same cell to cross "how do you launch Dask"
with "Do you use Dask as part of a larger group" with gradient defined by
rows.  It's an obvious result, but it clearly shows that solo users are
using SSH, while group-wise users have either HPC or Kubernetes deployments.

TomAugspurger:
Updated to point the binder URL in this post to point to the notebook that will be created by https://github.com/dask/dask-examples/pull/91

TomAugspurger:
I'll probably wait until tomorrow to publicize this out and the dask-examples binder with the survey notebook.

PR #48, opened 2019-08-05
Add draft of Dask + ITK post

mrocklin:


jni:
@mrocklin great post! I want to add a note for future exploration, since you mention GPUs: there is an OpenCL (:stuck_out_tongue_winking_eye:)-accelerated Richardson-Lucy implementation [here](https://github.com/maweigert/gputools/blob/master/gputools/deconv/deconv_rl.py).

jakirkham:
Thanks for working on this @mrocklin! Looks really good.

Submitted some minor fixes against your branch ( https://github.com/mrocklin/dask-blog/pull/1 ). Feel free to do what you like with these.

jakirkham:
At a higher level, one thing we discussed in the issue and haven't really handled here is the PSF may be acquired with a different scale. So we haven't handled rescaling in this post. I'm not sure it is critical for us to show how to do this in the post because users would still be able to do this in whatever way made sense previously (no need for Dask to get involved here). Though we could mention this somewhere. Maybe @rxist525 @thewtex have thoughts here?

_posts/2019-08-05-image-itk.md
```diff
@@ -0,0 +1,686 @@
...
+imgs = imgs.astype(np.float32)
+psf = psf.astype(np.float32)
+
+# Apply Richardson-Lucy Deconvolution¶
+def richardson_lucy_deconvolution(img, psf, iterations=1):
+    """ Apply deconvolution to a single chunk of data """
+    import itk
+
+    img = img.squeeze()  # remove leading two length-one dimensions
+    psf = psf.squeeze()  # remove leading two length-one dimensions
```
jakirkham: Just as a note, would personally avoid `.squeeze()`. Have seen it result in buggier code in practice.

_posts/2019-08-05-image-itk.md
```diff
@@ -0,0 +1,686 @@
...
+    provided, so we add these singleton dimensions back in
+
+    ```python
+    result = result[None, None, ...]  # Add back the leading length-one dimensions
+    ```
+
+    Again, this is straightforward for users who are accustomed to NumPy
+    slicing syntax, but does need to be done each time.
+    This adds some friction to our development process,
+    and is another step that can confuse users.
```
jakirkham, line 529: ITK might be able to support 5-D here (though @thewtex would be able to comment). However the binaries shipped support only a few different ranks of arrays. Also ITK is pretty good about noting this when one tries to make an image with an unsupported rank. This is because at the C++ level the array rank is templated (like a 2-D and a 3-D array are different types). ITK chooses to build out a few, but not all of these array types for various reasons (like to keep the final binary compact). Even if we were able to convert to a higher rank array, we may encounter other issues like the algorithm requires a specific rank. Though this likely has a clear error and would be no different from an error another library might raise. Not sure if or how we should mention this here, just wanted to raise awareness on this point.

_posts/2019-08-05-image-itk.md
mrocklin: Done.  Thanks.

_posts/2019-08-05-image-itk.md
mrocklin: I think I tried this and things didn't work correctly.  Should I try again @thewtex ?

mrocklin:
> @mrocklin great post! I want to add a note for future exploration, since you mention GPUs: there is an OpenCL (😜)-accelerated Richardson-Lucy implementation here.

Thanks for the pointer @jni .  I'm inclined to wait to discuss actual GPU implementations of deconvolution (I assume that there are a few) until some future post if that's alright.  I'll try to remember to ping you on that one as well so that you can keep me honest :)

mrocklin:
I plan to merge and release this tomorrow morning if there are no further objections.

_posts/2019-08-05-image-itk.md
jakirkham: Right I'm not saying the ITK binaries support this now. They were not built with this functionality. I'm saying one could build ITK in a way where this could work. This makes the binaries bigger (hence why it wasn't done according to the error message), but perhaps tolerable if you need it. Just want to make sure we are being fair to ITK here.

jakirkham:
Any thoughts on the [PSF comment above]( https://github.com/dask/dask-blog/pull/48#issuecomment-519059573 )?

VolkerH:
@jni pointed out this WIP blog post to me. 
At this stage the gputools RL implementation @jni linked to is very basic and more a proof-of-concept than something you would want to use in production. 

I just wanted to make you aware of this notebook 

https://github.com/hammerlab/flowdec/blob/master/python/examples/notebooks/Tile-by-tile%20deconvolution%20using%20dask.ipynb

where I used dask to break a deconvolution problem into chunks that fit into GPU RAM and then applying the excellent flowdec deconvolution by @eric-czech .

rxist525:
> At a higher level, one thing we discussed in the issue and haven't really handled here is the PSF may be acquired with a different scale. So we haven't handled rescaling in this post. I'm not sure it is critical for us to show how to do this in the post because users would still be able to do this in whatever way made sense previously (no need for Dask to get involved here). Though we could mention this somewhere. Maybe @rxist525 @thewtex have thoughts here?

I'm a bit confused about the PSF scale issue -- are we talking intensity scaling or voxel scaling?
In the first case of the intensity scaling, I think it'd be fair to assume the user is generating a PSF with sufficient amplitude to be able to get an appropriate OTF support from it for the deconvolution step. The alternative is to have an option for blind deconvolution.

mrocklin:
> Any thoughts on the PSF comment above?

I don't have any thoughts, no.  If it's something that should be added can you add it?

The blogpost as is gives a fairly cohesive message I think.  I'm inclined to publish it as is and move on and be productive elsewhere (we've been slow recently).  If someone has something that is important to say here though then please do so.  I don't know how important the PSF scaling issue is though.  Should it block publication?

rxist525:
It shouldn't. I think a small note/disclaimer that users are responsible
for generating a proper PSF.

jakirkham:
Thanks @rxist525! That answers my question. Let's go ahead and merge then.

thewtex:
@mrocklin @jakirkham outstanding post! :100: :1st_place_medal: :taco: Thank you so much for writing it up! It looks great to me.

Sorry, I was out traveling this week, and I missed all this wonderful action. Just a heads up, I will be traveling today, and back late Monday.

thewtex:
> Any thoughts on the PSF comment above?

> It shouldn't. I think a small note/disclaimer that users are responsible for generating a proper PSF.

Yes -- it is important that the PSF is sampled with the same voxel spacing as the image. If they do not, resampling could be applied. Ideally, the [images carry their spacing metadata](https://discourse.itk.org/t/images-in-physical-space-in-python/2124/16) so the function can detect and throw an exception when incompatible image and psf arguments are passed.

jakirkham:
Thanks for the feedback. Yeah it would be nice to include this note, but couldn't think of a good spot to add it. If you think of one, would be happy to review/merge a PR. 🙂

_posts/2019-08-05-image-itk.md
thewtex: Just to follow-up here: as @jakirkham mentions, ITK has compile-time optimizations based on the image dimension -- the Python packages are built for 2D and 3D dimensions. But, with the GUFunc support suggested by @mrocklin , these "core dimensions" will naturally be applied to images that have additional array dimensions that may correspond to channels or time.

thewtex:
> If you think of one, would be happy to review/merge a PR. slightly_smiling_face

Voila :-): https://github.com/dask/dask-blog/pull/51

PR #49, opened 2019-08-08
Add missing `DGX` import

jakirkham:
This example didn't run for me without this change.

### DIFF
--- a/_posts/2019-06-09-ucx-dgx.md
+++ b/_posts/2019-06-09-ucx-dgx.md
@@ -619,6 +619,8 @@ import cupy
 import dask.array as da
 from dask.distributed import Client, wait
 from distributed.utils import format_time, format_bytes
+from dask_cuda import DGX
+
 
 async def f():
 
--- a/_posts/2019-06-09-ucx-dgx.md
+++ b/_posts/2019-06-09-ucx-dgx.md
@@ -619,7 +619,7 @@ import cupy
 import dask.array as da
 from dask.distributed import Client, wait
 from distributed.utils import format_time, format_bytes
-from dask_cuda import LocalCUDACluster
+from dask_cuda import DGX
 
 
 async def f():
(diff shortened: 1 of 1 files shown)


jakirkham:
I guess it is also defined earlier in the blog. Anyways this is a useful piece of code to drop into a script and just run. It's nice if it just runs out of the box to cutdown on setup time.

jacobtomlinson:
Sorry this has sat here for so long @jakirkham. Is this change still relevant?

pentschev:
The `DGX` class has been removed as of dask-cuda 0.14. If updating is important, we should replace it by `LocalCUDACluster`.

_posts/2019-06-09-ucx-dgx.md
```diff
@@ -619,6 +619,8 @@ import cupy
 import dask.array as da
 from dask.distributed import Client, wait
 from distributed.utils import format_time, format_bytes
+from dask_cuda import DGX
```
jacobtomlinson, line 622:
```suggestion
from dask_cuda import LocalCUDACluster
```

jakirkham:
Except we are using `DGX` below.

jakirkham:
I looked at updating this to use `LocalCUDACluster`, but that turns into rewriting the blogpost as one of the examples shows the `DGX` code. If we need to specify a note that things have changed in newer versions, that seems fine. Though I'd prefer not to make too many changes here (it's a post from last year).

pentschev:
The `DGX` class was a specialization of `LocalCUDACluster`. All that specialization has been moved into `LocalCUDACluster`, so if you only replace `DGX` by `LocalCUDACluster`, that should suffice. I have no strong opinions on whether we should update this, so I'll leave the final decision to you.

jakirkham:
Yeah there is just a whole section including the [`DGX` code spelled out]( https://github.com/dask/dask-blog/blob/38ab575070abde27f1589a35f3303ef92b62ce97/_posts/2019-06-09-ucx-dgx.md#dgx ). So changing the imports doesn't make much sense in context.

I think it is ok that things are out-of-date (it's an old blog). If we wanted to update them, it would be better to do a new short blogpost that references this and notes how one would do things today.

jsignell:
Ok so it seems like this can safely be closed. @dask/maintenance

PR #50, opened 2019-08-09
WIP: Pairwise distance dask cuda blog post

alimanfoo:
Initial work towards a pairwise distance blog post, collaborating with @jakirkham.


mrocklin:
Thanks for doing this @alimanfoo and @jakirkham 

From our brief encounter on this last month, I also found the narrative from your numba.cuda.jit exploration motivating

1.  Wrote something on the CPU
2.  Tried numba.cuda.jit on a low-end GPU, wasn't any faster 
3.  Wrote it again, got a modest speedup
4.  Tried it on a V100, got 100x speedup
5.  Did a bunch of things (what were they?) got up to 10x speedup on laptop

_posts/2019-08-09-pairwise-distance.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "        \n",
+    "def pairwise_cityblock_cuda(x):\n",
+    "\n",
+    "    # Set up output array.\n",
+    "    n = x.shape[1]\n",
+    "    n_pairs = (n * (n - 1)) // 2\n",
+    "    out = cuda.device_array(n_pairs, dtype=np.float32)\n",
+    "\n",
+    "    # Let numba decide number of threads and blocks.\n",
+    "    kernel_spec = kernel_cityblock_cuda.forall(n_pairs)\n",
```
jakirkham: Here we use the raveled index to pass to `forall` (`n_pairs` is computed a few lines earlier).

_posts/2019-08-09-pairwise-distance.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "\n",
+    "@cuda.jit\n",
+    "def kernel_cityblock_cuda(x, out):\n",
+    "    m = x.shape[0]\n",
+    "    n = x.shape[1]\n",
+    "    n_pairs = (n * (n - 1)) // 2\n",
+    "    pair_index = cuda.grid(1)\n",
+    "    if pair_index < n_pairs:\n",
+    "        # Unpack the pair index to column indices.\n",
+    "        j, k = square_coords_cuda(pair_index, n)\n",
```
jakirkham: Here is where the unraveling is occurring for context.

jakirkham (commented):
Just noting this, @kkraus14 had suggested using `forall` as a way to avoid setting the threads to use. @alimanfoo found that this worked very nicely both from a user perspective and a performance perspective.

Though one thing that this use case highlights here is users will want to loop over several dimensions, which `forall` doesn't support currently (unless we are missing something?). As a result, the coordinates get raveled and then unraveled in the kernel.

If this pattern of raveling/unraveling with `forall` is reasonable, it would be handy to have this handled for the user (for example calling `.forall((n, n-1))`). Thus making it a bit easier to write Numba CUDA kernels. Thoughts @seibert? Happy to raise an issue to discuss this further if it is of interest.

_posts/2019-08-09-pairwise-distance.ipynb
```diff
@@ -0,0 +1,720 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def pairwise_cityblock_cpu(x):\n",
+    "    assert x.ndim == 2\n",
+    "    out = spd.pdist(x.T, metric='cityblock')\n",
```
jakirkham: To record a point from earlier for follow up, `pdist` is likely using `float64`. Though the kernels written below are `float32`. Would be good to standardize these a bit to have a fair comparison. That said, this will likely only be a 2x change relative to what we have here (unless I'm missing things).

One option would be to write a Numba kernel for this case as well. Though after talking with @alimanfoo, it seems SciPy's implementation is a bit more performant than a naive Numba kernel. Also SciPy seems to handle C and Fortran order arrays with identical performance. So there may be some other tricks buried in the SciPy implementation that would be worth learning from.

_posts/2019-08-09-pairwise-distance.ipynb
```diff
@@ -0,0 +1,720 @@
...
+    "x_big_dask_cuda = x_big_dask.map_blocks(cuda.to_device)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 26,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# launch a local cuda cluster?"
```
jakirkham: It seems that just running with multiple Dask workers using large data on a single GPU doesn't yield much here as the GPU is basically saturated with work. Though we suspect having multiple GPUs will be more useful as one can load chunks of data to compute on each GPU to get a speed up from parallelization.

alimanfoo:
Thanks @jakirkham. Just to xref https://github.com/numba/numba/issues/4309 in which I asked about another possible way to simplify running cuda kernels, including over multiple dimensions. As a user I would generally be happy with any approach that means I don't have to hard-code the number of threads-per-block. I realise that's maybe hard to do in all cases, but if there was a reasonable approximation it could remove an entry barrier to programming cuda kernels.

An interesting point of detail here as well, if the pairwise distance was run as a 2D grid, that would mean `n**2` threads get scheduled, but we only want to run `n choose 2` threads. It would be easy to add some logic so that a thread only runs if it is in the upper triangle, for example, but then is there any cost to scheduling around twice as many threads as are needed? Just a point of curiosity.

mrocklin:
I wonder if guvectorize can help here when parallelizing across additional
dimensions?  It seems like that might also handle some of the ambiguity
around threads

jakirkham:
> I wonder if guvectorize can help here when parallelizing across additional
dimensions?

Yeah we chatted about this as well. I think it would be a good thing to try.

It's worth noting that pairwise distance is raveling the results a bit. So there may be a trick that we didn't quite wrap our heads around. Certainly it makes sense for the non-pairwise distance case.

jakirkham:
> An interesting point of detail here as well, if the pairwise distance was run as a 2D grid, that would mean `n**2` threads get scheduled, but we only want to run `n choose 2` threads. It would be easy to add some logic so that a thread only runs if it is in the upper triangle, for example, but then is there any cost to scheduling around twice as many threads as are needed? Just a point of curiosity.

This is a good question. I'm not sure the answer. Hopefully one of the other CUDA experts cc'd can weigh in.

jakirkham:
Thanks for working on this a bit with me, @alimanfoo, and sharing your experience thus far. Tried to capture our conversation from earlier in the comments. Please feel free to fill in or correct anything as needed.

stuartarchibald:
> Just noting this, @kkraus14 had suggested using `forall` as a way to avoid setting the threads to use. @alimanfoo found that this worked very nicely both from a user perspective and a performance perspective.
> 
> Though one thing that this use case highlights here is users will want to loop over several dimensions, which `forall` doesn't support currently (unless we are missing something?). As a result, the coordinates get raveled and then unraveled in the kernel.
> 
> If this pattern of raveling/unraveling with `forall` is reasonable, it would be handy to have this handled for the user (for example calling `.forall((n, n-1))`). Thus making it a bit easier to write Numba CUDA kernels. Thoughts @siebert? Happy to raise an issue to discuss this further if it is of interest.

CC @seibert ^ assuming that's a typo and Stan is meant?

jakirkham:
> CC @seibert ^ assuming that's a typo and Stan is meant?

Yep, thanks for catching and correcting that Stuart! 😄

jakirkham:
I took the liberty of turning this into markdown and filling in some text. Hope that is ok @alimanfoo. Happy to revert, move around, and work through stuff here as you see fit. Also please feel free to overwrite changes I've made if you see this going a different direction or if I've missed things. The genomics bits in particular are pretty rough as I'm definitely **not** the expert there. 😄

jakirkham:
Also @mrocklin if you have other suggestions here, they would be welcome. 🙂

alimanfoo:
Thanks @jakirkham. I don't think I'll get to this before going on leave, but if you don't mind hanging on to this for a couple of weeks I'd be happy to help finish it up when back. 

Also happy to add a few notes on things that I tried in earlier experiments that didn't work so well as suggested by @mrocklin if you think it would be worth it (basically it was trying the kernel parallelising over rows instead of pairs, which then required an atomic add, which I think slows it down).

jakirkham:
Yep, no worries. Just wanted to fill in some things so we have an idea of what is here and what to do when you are back. Figured it might make things more approachable when we return.

That could be interesting. As I filled this out, there were some points in the narrative that could benefit from these earlier parts of the story. Like how did you implement the distance function before using `forall`? Also how well did that work? Were there any tricks that others should be aware of?

alimanfoo:
>
> That could be interesting. As I filled this out, there were some points in
> the narrative that could benefit from these earlier parts of the story.
> Like how did you implement the distance function before using forall?
>
Yep, could add that too. I wondered if it might be good to add some of
these things in as a kind of post-script, something like "other things we
tried", so the main body of the post is relatively concise, but there are
some further details for the interested reader. Happy to do it either way.

jakirkham:
Yeah there was a sizable bit near the end that I condensed down to a small bit of code and a few lines of text. There is a balance to be struck between info and length. Not sure what the right balance is in this case, but think we will be able to figure it out. My hope is we are kind of close with the content here.

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+---
+{% include JB/setup %}
+
+In large scale genomics analysis, scientists try to reduce sequencing data from
+multiple individuals into groups that share common traits. This requires
+measuring how sequences differ from each other. A common way to do this is to
+use a distance metric like [city block](
+https://en.wikipedia.org/wiki/Taxicab_geometry ) or others. As this metric ends
+up being computed pairwise over all of the samples, this can be a real
+bottleneck. Thus it is pairmount to have a performant distance computation
```
stuartarchibald:
```suggestion
bottleneck. Therefore it is paramount to have a performant distance computation
```

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+draft: true
+theme: twitter
+---
+{% include JB/setup %}
+
+In large scale genomics analysis, scientists try to reduce sequencing data from
+multiple individuals into groups that share common traits. This requires
+measuring how sequences differ from each other. A common way to do this is to
+use a distance metric like [city block](
+https://en.wikipedia.org/wiki/Taxicab_geometry ) or others. As this metric ends
```
stuartarchibald:
```suggestion
https://en.wikipedia.org/wiki/Taxicab_geometry ). As this metric ends
```

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+up being computed pairwise over all of the samples, this can be a real
+bottleneck. Thus it is pairmount to have a performant distance computation
+implementation to really leverage the hardware.
```
stuartarchibald:
```suggestion
implementation that makes full use of the hardware available.
```

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+
+
+TODO Might be nice to have a picture or graph of what some data looks like
+
+TODO explain what the data mean.
+
+So a representative dataset, might look something like this.
```
stuartarchibald:
```suggestion
A representative dataset, might look something like this:
```

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+
+```python
+# simulate some genetic data
+x = np.random.choice(np.array([0, 1, 2], dtype='i1'),
+                     p=[.7, .2, .1,],
+                     size=(20_000, 1_000))
+```
+
+To start out, one might leverage a distance implementation from the highly
+optimized SciPy library. This is quick and can be dropped with relatively
```
stuartarchibald:
```suggestion
optimized SciPy library. This is quick and can be employed with relatively
```

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+little effort as long as one's data fits what the function expects. So one
```
stuartarchibald:
```suggestion
little effort as long as one's data fits what the function expects. Consequently one
```

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+</svg>
+</td>
+</tr>
+</table>
+
+
+With our Dask Array in hand, we can write a simple function to facilitate
+applying our distance function on the data. By doing a little bit of work to
+keep things general, we can easily swap in different distance functions based
+on the computation we need to do or leverage more performant implementations
```
stuartarchibald, line 112: There is a lot of `leverage` in this text, perhaps consider synonyms such as {engage, use, exercise, employ, exploit}?

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+```python
+pairwise_cityblock_dask(x_dask, f=pairwise_cityblock_cpu).compute()
+```
+
+With this, we are able to get more of a speedup. We did a little bit of work,
+but not so much. Also we have managed to leverage the distance function from
+our serial case without issues.
+
+What if we still want more performance than even this offers? Well if we have a
+GPU in our computer or a cluster of them, we could leverage that compute
```
stuartarchibald:
```suggestion
GPU in our computer, or a cluster of them, we could leverage that compute
```

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+resource for our problem.
+
+First we need to move our data to the GPU. There are several ways we could do
+this. Though one reasonably approachable way is to use Numba. We could write
```
stuartarchibald:
```suggestion
this, though one reasonably approachable way is to use Numba. We could write
```

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+the following to handle our in-memory case.
+
+```python
+x_cuda = cuda.to_device(x)
```
stuartarchibald, line 158: Should explain what this does?

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+
+We can also handle the Dask case pretty easily by simply writing this.
+
+```python
+x_dask_cuda = da.map_blocks(cuda.to_device, x)
+```
+
+This may take a moment as data moves from the CPU to the GPU.
+
+At a first pass, we might write something Numba's CUDA JIT. Here we might need
```
stuartarchibald, line 169: I'm not sure this sentence makes sense? Also might be a good idea to add a one sentence on "what a JIT is".

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+to tune how many threads we use per block.
+
+
+TODO Add this code
+
+
+Alternatively we could be a bit clever and write this code using `forall`. This is pretty handy as we now no longer need to think about CUDA threads per block. There is a caveat though; we need to collapse our `for`-loops into 1 loop. One can do this with a bit of ingenuity.
```
stuartarchibald:
```suggestion
Alternatively we could be a bit clever and write this code using `forall`. This is pretty handy as we now no longer need to think about CUDA threads per block. There is a caveat though; we need to collapse our `for`-loops into one loop. One can do this with a bit of ingenuity.
```
also this needs a line wrap.

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+We can then try to run this on our representative dataset. This does reasonably
+well and we haven't had to think to hard about the implementation. In some
+cases, this may already be sufficient.
+
+However we may find that while the performance is reasonable, we have a lot of
+samples to process. We could speed things up a bit by running through small
+batches of data in parallel. An easy way to get started on this work would be
+to make a Dask Array that handles operating on our data.
```
stuartarchibald: Link `Dask Array` to docs?

stuartarchibald:
```suggestion
to make a Dask Array that handles operations on our data.
```

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+Now we can write something like this.
+
+```python
+pairwise_cityblock_dask(x_dask, f=pairwise_cityblock_cpu).compute()
+```
+
+With this, we are able to get more of a speedup. We did a little bit of work,
+but not so much. Also we have managed to leverage the distance function from
```
stuartarchibald:
```suggestion
but not a great amount. Also we have managed to leverage the distance function from
```

stuartarchibald: "leverage"

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+our serial case without issues.
+
+What if we still want more performance than even this offers? Well if we have a
```
stuartarchibald:
```suggestion
What if we still want more performance than this offers? Well, if we have a
```

_posts/2019-08-09-pairwise-distance.md
stuartarchibald: "leverage" perhaps "use"?

_posts/2019-08-09-pairwise-distance.md
stuartarchibald: Add link to Numba?

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+GPU in our computer or a cluster of them, we could leverage that compute
+resource for our problem.
+
+First we need to move our data to the GPU. There are several ways we could do
+this. Though one reasonably approachable way is to use Numba. We could write
+the following to handle our in-memory case.
```
stuartarchibald, line 155: In memory with respect to what? In machine RAM vs in GPU RAM.

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+    n = np.float32(n)
+    j = (((2 * n) - 1) - math.sqrt((1 - (2 * n)) ** 2 - (8 * pair_index))) // 2
+    k = pair_index - (j * ((2 * n) - j - 1) / 2) + j + 1
+    j = np.int64(j)
+    k = np.int64(k)
+    return j, k
+
+
+@cuda.jit
+def kernel_cityblock_cuda(x, out):
```
stuartarchibald, line 194: I feel it would be useful to have a version of the CPU equivalent algorithm for comparison as it's not obvious the level of effort required to convert from CPU to CUDA as the current baseline CPU version is a SciPy function call (`scipy.spatial.distance.spd`)? A useful side effect of this would be a `nopython` mode JIT compiled variant which gives a hint at CPU/parallel CPU performance vs. CUDA.

A further side effect would be that it might be help in working out a strategy for conversion of such algs. `njit(cpu) -> guvectorize(cpu) -> guvectorize(cuda) -> cuda kernels -> optimising cuda kernels` (mixed in with dask?)?

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+
+
+Alternatively we could be a bit clever and write this code using `forall`. This is pretty handy as we now no longer need to think about CUDA threads per block. There is a caveat though; we need to collapse our `for`-loops into 1 loop. One can do this with a bit of ingenuity.
+
+```python
+import math
+
+
+@cuda.jit(device=True)
+def square_coords_cuda(pair_index, n):
```
stuartarchibald, line 183: Some LaTeX spelling out the maths and/or a wiki link to a canonical form of this alg would be beneficial.

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+    
+    return out
+```
+
+As with any JIT, we benefit by doing a warmup run first. Though that's easy
+enough to do. Then we are ready to compute our result.
+
+```python
+dist_cuda = pairwise_cityblock_cuda(x_cuda)
+cuda.synchronize()
```
stuartarchibald, line 234: Might want to explain why this is needed, it's a common stumbling block in assessing the performance of CUDA functions (Numba core devs answer this question regularly)?

_posts/2019-08-09-pairwise-distance.md
```diff
@@ -0,0 +1,262 @@
...
+
+cluster = LocalCUDACluster()
+client = Client(cluster)
+```
+
+Then distribute our computation over that cluster.
+
+```
+dist_big_cuda = pairwise_cityblock_dask(x_big_dask_cuda, f=pairwise_cityblock_cuda).compute()
+```
```
stuartarchibald, line 262: It'd probably be worth making this into a notebook so that it can be downloaded and executed, it'd also shake out any bugs in the code blocks in the document? Having some canned/expected output would definitely help with ensuring that algorithm translations etc are done correctly.

stuartarchibald (commented):
Thanks for writing this, I've read through and think it's a good self contained use case for demonstrating how to go from a CPU alg to a potentially distributed GPU alg. The narrative does a good job of explaining a method to perform such changes.

I've made a mixture of suggestions (acceptance is entirely optional :)) categorised approximately as follows:

 * stylistic/wording
 * ways to reinforce the narrative/motivation
 * additional items to help develop the CPU->GPU strategy described within

There's also the question of determining whether a kernel and the launch config is making good use of the hardware. Perhaps that should/could be a follow-up and could contain more code tweaks and exploration of performance analysis tooling?

jakirkham:
@alimanfoo , would you have a chance to go through this and fill in any scientific background I may have lost? Also feel free to change any of the text as you see fit. 🙂

alimanfoo:
@jakirkham apologies, I've been totally buried. Will try to get back to this asap. ccing @jeromekelleher for interest.

_posts/2019-08-09-pairwise-distance.md
jakirkham:
```suggestion
to make a [Dask Array]( https://docs.dask.org/en/latest/array.html ) that handles operations on our data.
```

_posts/2019-08-09-pairwise-distance.md
jakirkham:
```suggestion
GPU in our computer or a cluster of them, we could use that compute
```

_posts/2019-08-09-pairwise-distance.md
jakirkham:
```suggestion
this; though, one reasonable way is to use [Numba]( https://numba.pydata.org ). We could write
```

_posts/2019-08-09-pairwise-distance.md
jakirkham:
```suggestion
GPU in our computer, or a cluster of them, we could use that compute
```

_posts/2019-08-09-pairwise-distance.md
jakirkham: Handled in latest commit.

_posts/2019-08-09-pairwise-distance.md
jakirkham: Yeah this started as a notebook originally. Then we moved it into this blogpost form so we could publish and share it a bit more broadly.

jacobtomlinson:
It has been a couple of years since there has been activity here. Thanks for all the effort that went into this but I'm going to close it out for now in an effort to tidy things here a little. It would be great to see more submissions like this though.

PR #54, opened 2019-09-10
Add Jupyter on Dask post

mrocklin:
cc @bollwyvl @yuvipanda @mmccarty 

I had a conversation about this with @mmccarty , then asked a question of @bollwyvl , and saw that it was easy.  I figured I'd write it up.  I only spent like 30m on this, so may not do a ton of editing here, but if people have suggestions on how this can be improved they would be welcome.


mmccarty:
Wow, this is great!  Thanks for figuring this out.  I'm going to do some data splunking tomorrow, so this will be quite useful.

_posts/2019-09-10-jupyter-on-dask.md
```diff
@@ -0,0 +1,94 @@
...
+---
+{% include JB/setup %}
+
+If you want, you can have Dask set up a Jupyter notebook server for you,
+co-located with the Dask scheduler.  There are many ways to do this, but this
+blog post lists two.
+
+### First, why would you do this?
+
+Sometimes people inside of large institutions have complex deployment pains.
```
mmccarty: complex deployment ~~pain~~ plans

Also, institutions may have a large number of distinct, isolated cloud accounts where data cannot be moved.  We need to be able to quickly bring in ephemeral compute infrastructure from the ground up.  Perhaps for ad-hoc data analysis without the need to on board to a complex enterprise solution that could take months to implement.

_posts/2019-09-10-jupyter-on-dask.md
```diff
@@ -0,0 +1,94 @@
...
+```yaml
+distributed:
+  scheduler:
+    preload: ["/path/to/jupyter-preload.py"]
+```
+
+### Really though, you should use something else
+
+This is mostly a hack.  If you're at an institution then you should ask for
+something like [Juptyer Hub](https://jupyter.org/hub).
```
bollwyvl: `JupyterHub`

_posts/2019-09-10-jupyter-on-dask.md
```diff
@@ -0,0 +1,94 @@
...
+    app.initialize([])  # add command line args here if you want
+
+client.run_on_scheduler(start_jupyter_server)
+```
+
+If you have a complex networking setup (maybe you're on the cloud or HPC and
+had to open up a port explicitly) then you might want to install
+[jupyter-server-proxy](https://jupyter-server-proxy.readthedocs.io/en/latest/)
+(which Dask also uses by default if installed), and then go to
+[http://scheduler-address:8787/proxy/8888](https://example.com) .  The Dask dashboard can route your
```
jacobtomlinson: It's funny to see the url this way round. We've closed the loop here so I expect we could have `:8787/proxy/8888/proxy/8787/proxy/8888` 😆.

mrocklin: :)

PR #57, opened 2019-09-30
MAINT: Small image/video width edits

stsievert:
**What does this PR implement?**
It makes sure the post isn't too wide on narrow screens. It does this by setting the `max-width` attribute of the `img`/`video` tags.

I have tested this PR to make sure it has the correct behavior.


mrocklin:
Thanks @stsievert !

PR #65, opened 2020-07-04
Add draft of high performance scheduling

mrocklin:


mrocklin:
This is rough, I'm just spewing out some thoughts for now

jacobtomlinson:
Interesting thoughts!

> Also, I suspect that splitting out also probably means that we'll probably lose the dashboard, unless those other groups are very careful to expose the same state to Bokeh.

I wonder if there would be value in decoupling the dashboard from the scheduler. If the dashboard were a standalone bokeh application which had access to state via Dask communication it would be available to any scheduler implementation.

mrocklin:
> I wonder if there would be value in decoupling the dashboard from the scheduler. If the dashboard were a standalone bokeh application which had access to state via Dask communication it would be available to any scheduler implementation.

Having direct access to the scheduler state has been very valuable for making the dashboard evolve quickly.  There are significant costs to decoupling the two.  It's definitely possible though.

quasiben:
Thanks for writing this up!

jacobtomlinson:
> There are significant costs to decoupling the two.

Perhaps there are things that can be done in the communication to reduce this cost?

mrocklin:
I meant costs in terms of human development cost.  Currently it feels luxurious to be able to just reach in and touch the entire scheduler state and plot whatever we want.  Not having an interface boundary here is nice.

We can have an interface boundary though.  Probably this means that we would want to host a bunch of websocket endpoints?

_posts/2020-07-04-high-performance-scheduling.md
```diff
@@ -0,0 +1,327 @@
...
+
+Who cares?
+----------
+
+First, this is a problem that affects about 1% of Dask users.  These are people
+who want to process millions of tasks relatively quickly.  Let's list a few use
+cases:
+
+1.  Xarray/Pangeo workloads on the 100TB scale
+2.  NVIDIA benchmarking (GPUs make computing fast, so other costs become
```
kkraus14: I'd like to clarify that it isn't just benchmarking. We have multiple customer workloads bottlenecked by the task graph handling.

jacobtomlinson:
> We can have an interface boundary though. Probably this means that we would want to host a bunch of websocket endpoints?

I had imagined using the RPC or extending the current websocket plugin for this. It would result in one RPC or websocket connection between the dashboard server and the scheduler. 

Currently the websocket is push only, all plugin events are serialised into JSON and pushed down the socket. All RPC connections are pull only, the dashboard would poll the state of the scheduler.

I think the dashboard uses a mix of both methods, so one or the other would need to be extended to fill in the gap.

_posts/2020-07-04-high-performance-scheduling.md
```diff
@@ -0,0 +1,344 @@
...
+Getting culling to work properly may require us to teach each of the individual
+graph layers how to track dependencies in each layer type and cull themselves.
+This may get tricky.
+
+Slicing is doable, we just need someone to go in, grok all of the current
+slicing optimizations, and make high level graph layers for these
+computations.  This would be a great project for a sharp masters student
+
+
+#### Send speculative tasks to the workers
```
TomAugspurger: I don't understand the change this proposal is making. Is it that when task `B` depends on task `A`, we assign task `B` to a machine before `A` is finished? And this is helpful because... (this is the part I don't get)

_posts/2020-07-04-high-performance-scheduling.md
```diff
@@ -0,0 +1,344 @@
...
+2.  But this breaks low level graph optimizations, fuse, cull, and slice fusion
+    in particular.  We can make these unnecessary with two changes:
+    -   We can make high level graphs considerably smarter to handle cull and slice fusion
+    -   We can move a bit more of the scheduling down to the workers to remove
+        the advantages of low-level fusion
+3.  Then, once all of the graph manipulation happens on the scheduler, let's
+    try to accelerate it, hopefully in a language that the current dev
+    community can understand, like Cython
+4.  At the same time in parallel, let's take a look at our network stack
+
```
TomAugspurger: 5. Provide more diagnostics on the dashboard for stages 3-6, which might be less work :)

This doesn't help with the first couple stages though.

TomAugspurger (commented):
Overall I'm on board with this (laying out these problems, exploring the proposed solutions, and finding time to implement them). Maintainability is my primary concern but the likely benefits seem to outweigh the costs. Expanding HLGs will make the dask collections harder to pick up, but if we get the abstraction right hopefully it won't be too much harder. Likewise for rewriting (parts of) the schedulers.

But serving the users with high demands has been good for the project so far. I'd hate for them to have to go elsewhere.

_posts/2020-07-04-high-performance-scheduling.md
mrocklin: I would like to remove the call to the low-level `fuse` optimization function.  However, in many workloads scheduling is better if I/O operations like `read_parquet` are immediately followed by the next sequence of tasks, like `.x.sum()` rather than waiting for those tasks to come from the scheduler.  (usually the worker starts on some other `read_parquet` task and fills up memory a bit faster otherwise).  

By having the subsequent tasks already on the worker we get the same advantages of low-level task fusion, but without actually having to fuse the tasks.

As an additional advantage, we'll separately account for costs of `read_parquet` and `add` and `sum` rather than have task prefixes like `read_parquet-add-sum`.

_posts/2020-07-04-high-performance-scheduling.md
mrocklin: I'm curious about your thoughts on this.  In general I agree that visibility over this process would help us improve it.  Historically this has been hard because some of these stages run on the client side.  Now that we're moving more of it over to the scheduler hopefully it's more possible.  

Would this be just an bar/pie chart of costs, or something else?

TomAugspurger: Off the top of my head, `<thing>.compute()` sends small messages to the scheduler with status updates

- "compute on `<thing>` requested"
- "optimized `<thing>` to ... tasks. Here it comes"
- "Sent `<thing>` in ...s"

And the dashboard represents these somehow (perhaps as a coarse progress bar).

But, these are sufficiently different from the broader goals in this post that I think we don't need to go into detail, and perhaps we ignore the suggestion entirely, if we think it distracts from the message.

mrocklin: @jacobtomlinson  tried this in a dask/distributed PR.  It's tricky because these calls don't currently await within them reliably.  For example `client.submit` doesn't need an await.  So it's hard to send something up to the scheduler before we start serializing and such.  

Hopefully though a lot of this pain goes away once the early stages are all cheap, and most of the administrative cost occurs in the scheduler.

jacobtomlinson: That was here if you want to see the conversation dask/distributed#3363

PR #69, opened 2020-07-22
Add GitHub Pages refresh workflow to rebuild the blog daily

jacobtomlinson:
This GitHub Actions workflow will cause the GitHub Pages site for the Dask Blog to be rebuilt every day at 3am UTC.

Currently, the site only gets built when something is merged into the `gh-pages` branch.

This workflow would enable us to future date blog posts, which will not get built into the site until that date (Jekyll does this automatically). By triggering the build daily we can schedule blog posts to go out in the future simply by future dating them and merging ahead of time.

cc @jsignell @mrocklin

### DIFF
--- a/.github/workflows/refresh.yml
+++ b/.github/workflows/refresh.yml
@@ -1,0 +1,17 @@
+name: Refresh
+
+on:
+  schedule:
+    - cron: "0 3 * * *" # Runs every day at 3am
+
+jobs:
+  refresh:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger GitHub pages rebuild
+        run: |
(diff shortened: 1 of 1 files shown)


jsignell:
Oh I like this idea a lot! Can you include a change to the readme that explains what the new protocol is for merging a blog post?

mrocklin:
I think that we can still merge blogposts as normal, but just with dates in
the future.  Jekyll only builds those posts that are in the past.

jsignell:
Oh I see. So the change in this PR is that we build every night instead of only building on merge. That sounds good to me!

jsignell (approved):

PR #77, opened 2020-10-21
Deconvolution post

jakirkham:
**Note:** This is still very much WIP at this stage.

A blogpost about using Richardson-Lucy deconvolution with Dask to process data larger than memory. Here we show both CPU and GPU results using the same implementation (thanks to [NEP-18]( https://numpy.org/neps/nep-0018-array-function-protocol.html )).

cc @quasiben


_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,347 @@
...
+
+[Last year](https://blog.dask.org/2019/08/09/image-itk) we experimented with Dask/ITK/Scikit-Image
+to perform large scale image analysis on a stack of 3D images.  Specifically, we looked at `deconvolution`
+a common method to _deblur_ images.  Now, a year later, we return to these experiments with a better
+understanding of how Dask and CuPy can interact, enhanced serialization methods, and support from the
+open-source community.
+
+Previously we used the [Richardson Lucy (RL)](https://en.wikipedia.org/wiki/Richardson%E2%80%93Lucy_deconvolution) deconvolution
+algorithm from ITK and [Scikit-Image](https://github.com/scikit-image/scikit-image/blob/master/skimage/restoration/deconvolution.py#L329).
+We left off at theorizing how GPUs could potential help accelerate these workflows.  Starting with Scikit-Image's
```
TomAugspurger:
```suggestion
We left off at theorizing how GPUs could potentially help accelerate these workflows.  Starting with Scikit-Image's
```

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,347 @@
...
+implementation, we naively tried replacing `scipy.signal.convolve` calls with `cupyx.scipy.ndimage.convolve`, and while
+performance improved, it did not improve _significantly_ -- that is, we did not get the 100X speed we were looking for
+
+Often, in working with images, we hit performance roadblocks in real space where data represents position.  When this happens,
+we can reach for a common tool, the [Fast-Fourirer Transform FFT](https://en.wikipedia.org/wiki/Fast_Fourier_transform).  FFTs convert
+the data to from real space to frequency space where data now represents the rate of change of intensity between pixels.  This conversion
+is extremely fast on both CPUs and GPUs and the algorithm we can write with FFTs similarly are accelerated.
+And while it is challenging to write an equivalent algorithm in FFT space (thank you mathematicians!)
+the cost of transformation + the cost of the algorithm is still significantly faster than performing the original algorithm in real space.
```
TomAugspurger: A bit of a mixed metaphor here. "the *cost* of ... is significantly *faster*".

Cost should be less / lower, or the speed should be faster.

quasiben: Thanks @TomAugspurger for the early review.  I'll fix these up and some other thoughts shortly

_posts/2020-07-30-image-redux.md
```diff
@@ -0,0 +1,474 @@
+---
```
GenevieveBuckley: Is it possible this file `_posts/2020-07-30-image-redux.md ` has sneaked in here by mistake? I don't think it should get lumped in with this pull request.

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,387 @@
...
+understanding of how Dask and CuPy can interact, enhanced serialization methods, and support from the
+open-source community.
+
+Previously we used the [Richardson Lucy (RL)](https://en.wikipedia.org/wiki/Richardson%E2%80%93Lucy_deconvolution) deconvolution
+algorithm from ITK and [Scikit-Image](https://github.com/scikit-image/scikit-image/blob/master/skimage/restoration/deconvolution.py#L329).
+We left off at theorizing how GPUs could potentially help accelerate these workflows.  Starting with Scikit-Image's
+implementation, we naively tried replacing `scipy.signal.convolve` calls with `cupyx.scipy.ndimage.convolve`, and while
+performance improved, it did not improve _significantly_ -- that is, we did not get the 100X speed we were looking for.
+
+Often, in working with images, we hit performance roadblocks in real space where data represents position.  When this happens,
```
GenevieveBuckley: I know this is still WIP and subject to change, but I think this paragraph is a bit confusing for someone not already familiar with the concepts. We can talk about this at a more appropriate time though.

quasiben:
Yup that's my fault.  I'll remove

_posts/2020-07-30-image-redux.md
jakirkham: Yeah it looks like this got added in commit ( https://github.com/dask/dask-blog/pull/77/commits/86ed738578f5f3981990759d1acbadd62003ad34 ). @quasiben is it ok if we just delete this file or is there some content that should be moved to `_posts/2020-10-23-deconvolution.md`?

GenevieveBuckley:
The SciPy Japan sprints are next week. Maybe it would be good to suggest someone turns this blogpost into a PR for deconvolution in dask-image.

Or I could *not* do that, if you'd prefer I hold off (or maybe you want to do that yourself @jakirkham). Thoughts?

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,387 @@
...
+[Last year](https://blog.dask.org/2019/08/09/image-itk) we experimented with Dask/ITK/Scikit-Image
+to perform large scale image analysis on a stack of 3D images.  Specifically, we looked at `deconvolution`,
+a common method to _deblur_ images.  Now, a year later, we return to these experiments with a better
+understanding of how Dask and CuPy can interact, enhanced serialization methods, and support from the
+open-source community.
+
+Previously we used the [Richardson Lucy (RL)](https://en.wikipedia.org/wiki/Richardson%E2%80%93Lucy_deconvolution) deconvolution
+algorithm from ITK and [Scikit-Image](https://github.com/scikit-image/scikit-image/blob/master/skimage/restoration/deconvolution.py#L329).
+We left off at theorizing how GPUs could potentially help accelerate these workflows.  Starting with Scikit-Image's
+implementation, we naively tried replacing `scipy.signal.convolve` calls with `cupyx.scipy.ndimage.convolve`, and while
```
stsievert: Why didn't `scipy.signal.convolve` work? It has a `method` parameter that allows switching between the FFT and direct methods of convolution. It defaults to `method="auto"` but it looks like with `method="fft"`, that function eventually [calls rfftn and irfftn][2] from [SciPy's FFT module][1], ~~which appears to have [NEP-18 support][3].~~

[1]:https://docs.scipy.org/doc/scipy/reference/fft.html
[2]:https://github.com/scipy/scipy/blob/9d60a26ed35dbab302daf4e23545f29a56a6d453/scipy/signal/signaltools.py#L497
[3]:https://github.com/scipy/scipy/blob/9d60a26ed35dbab302daf4e23545f29a56a6d453/scipy/fft/_basic.py#L1012

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,387 @@
...
+    <img src="/images/deconvolve.png" width="100%"></a>
+
+IMAGE DESCRIPTION
+
+Napari + Remote GPUs
+--------------------
+
+Deconvolution is just one operation and one tool, an image scientist or microscopist will need.  They will need other tools to help as the push to understand
+more of the underlying biology.  As a first step though, they will need tools to visualize the data. [Napari](https://napari.org/) is a multi-dimensional image viewer popular in the PyData Bio ecosystem.  As an experiment, we extended the demo above to not only work with Napari, but to also work on remote GPUs provided
+by [coiled.io](https://coiled.io/) [Example Notebook](https://gist.github.com/quasiben/29901204b20421b18f029e1fd07ed0cf)
```
GenevieveBuckley: It might be good to add a line saying that you could use other remote services like X or Y, but we picked Coiled because $reason

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,387 @@
...
+```
+
+With Dask and multiple GPUs, we measured deconvolution of an 8GB image in ~10 seconds! But this is just the first step towards accelerated image science.
+
+<a href="/images/deconvolve.png">
+    <img src="/images/deconvolve.png" width="100%"></a>
+
+IMAGE DESCRIPTION
+
+Napari + Remote GPUs
```
GenevieveBuckley: Very cool!

_posts/2020-10-23-deconvolution.md
stsievert: And does `cupyx.scipy.ndimage.convolve` use the direct method or FFT method of convolution?

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,387 @@
...
+</table>
+
+
+What we now have is a Dask array composed of 16 CuPy blocks of data.  Notice how Dask provides nice typing information in the SVG output.  When we moved from NumPy to CuPy, the block diagram above displays `Type: cupy.ndarray` -- these are especially helpful and remind us we are operating on data which lives on the GPU.
+
+The last piece we need before running the deconvolution is the PSF which should also be loaded onto the GPU:
+
+```python
+avg_psf = skimage.io.imread("/public/NVMICROSCOPY/AVG_PSF.tif")
+c_avg_psf = cp.asarray(avg_psf)
```
stsievert: What's the shape of PSF? That has a pretty drastic effect on the performance of direct convolution IIRC.

_posts/2020-10-23-deconvolution.md
quasiben: `cupyx.scipy.ndimage.convolve`, I believe, uses the direct method.  

When we started, there were no implementations of [NEP-35](https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html) nor [NEP-37](https://numpy.org/neps/nep-0037-array-module.html) which makes the first line of the definition for `scipy.signal.convolve`: `np.asarray(...)`, challenging.

_posts/2020-10-23-deconvolution.md
quasiben: The avg_psf is (128, 128).  My understanding is that these can be provided by the instrument manufacture.  Having said that, I'm not sure where this one came from -- we did, however, experiment with another PSF which is/was (128, 128, 128)

_posts/2020-10-23-deconvolution.md
quasiben: I agree.  I was trying to find some language for those unfamiliar with FFTs to help understand our motivations.  Do you think we should remove entirely or try to re-work?

_posts/2020-10-23-deconvolution.md
stsievert: > which makes the first line of the definition for scipy.signal.convolve: np.asarray(...), challenging.

Whoops, I see now. Sorry 'bout that.

Do you think SciPy's convolve would support an efficient implementation of FFTs on CuPy arrays now? Or would there by other challenges like `np.asarray`?

quasiben: > Whoops, I see now. Sorry 'bout that.

No worries at all

> Do you think SciPy's convolve would support an efficient implementation of FFTs on CuPy arrays now? Or would there by other challenges like np.asarray?

I think this is the biggest stumbling block for interoperability.  This also came up in a dask issue: https://github.com/dask/dask/issues/6440

jakirkham:
> The SciPy Japan sprints are next week. Maybe it would be good to suggest someone turns this blogpost into a PR for deconvolution in dask-image.
> 
> Or I could _not_ do that, if you'd prefer I hold off (or maybe you want to do that yourself @jakirkham). Thoughts?

Please go ahead 😄 No need to hold up things on my behalf 😉

_posts/2020-10-23-deconvolution.md
jakirkham: Using other services seems fine. My guess is any cloud or local resource would work. There's some setup effort incurred that varies depending on the scale, location, and what might already have been done to deploy there. Should we say something like this?

I don't know that we had a specific reason. It was easy. We were able to bug Matt to help us with things 😉. Is that worth mentioning? 🙂

_posts/2020-10-23-deconvolution.md
jakirkham: Scott, I'm guessing you are asking has something like PR ( https://github.com/scipy/scipy/pull/5608 ) been done for CuPy? If so, no, but there's an open issue ( https://github.com/cupy/cupy/issues/3378 ). It would certainly be great if something like that were implemented for CuPy 😉

As to leveraging existing work in SciPy, that would be nice. We brought this up last year ( https://github.com/scipy/scipy/issues/10204 ). The response was let's try this with one piece (FFTs) and see how that goes.  This got implemented in PR ( https://github.com/scipy/scipy/pull/10383 ), but requires a bit more hand-holding to dispatch effectively. Another question that would come up is at what level in the code would dispatching yield reasonable performance while also reusing as much code as possible. I don't know the answer there, but maybe someone else does.

In Dask-image Genevieve did a bunch of work on adding dispatching recently ( https://github.com/dask/dask-image/pull/151 ). Though this is a Dask-image solution to this problem. Perhaps if dispatching in SciPy were implemented generally, Dask-image could switch to that solution.

_posts/2020-07-30-image-redux.md
jakirkham: This has since been removed

_posts/2020-10-23-deconvolution.md
jakirkham: I tried rewording this and filling in other options. Please let us know what you think 😄

_posts/2020-10-23-deconvolution.md
jakirkham: Switched to "lower" in recent changes

_posts/2020-10-23-deconvolution.md
jakirkham: I tried giving an (admittedly hand wavy) explanation that didn't rely on understanding FFTs as much. Please let us know what you think :)

_posts/2020-10-23-deconvolution.md
stsievert: > I'm guessing you are asking has something like PR ( scipy/scipy#5608 )

Almost! I started that PR because I ran into similar speed issues with RL deconvolution that I resolved in https://github.com/scikit-image/scikit-image/pull/1792 (then I merged that upstream in https://github.com/scipy/scipy/pull/5608).

> requires a bit more hand-holding to dispatch effectively.

If I'm reading https://github.com/scipy/scipy/pull/10383 and https://github.com/scipy/scipy/issues/12652#issuecomment-667348345 correctly, it sounds like the extra bit of hand-holding would be a context manager. Is that correct? That doesn't sound too bad to me. So then the workflow below would work smoothly?

``` python
import cupy
import scipy.fft
from scipy.signal import convolve
import cupyx.scipy.fft as cp_fft
from skimage.restoration import richardson_lucy

img = cupy.arange(256**2).reshape((256, 256)).astype(float)
psf = cupy.arange(128**2).reshape((128, 128)).astype(float)

with scipy.fft.set_backend(cp_fft):
    out = convolve(img, psf, method="fft")

    # I'm especially interested in this:
    img_deblurred = richardson_lucy(img, psf=psf)
```

jakirkham: Ah ok. Well if you'd like to add any missing context to issue ( https://github.com/cupy/cupy/issues/3378 ), that would be great! I think it would be nice if we did similar things in CuPy.

That's a good question. Wasn't sure. So took it for a spin. Guessing `convolve` here means `from scipy.signal import convolve`? Unfortunately there's [`np.asarray` calls]( https://github.com/scipy/scipy/blob/79e64c140a518bc85d9105718301509f43e51444/scipy/signal/signaltools.py#L1375-L1376 ), which trips up CuPy. I'm not totally sure how we would back that out. Maybe we could check if the objects support `__array_function__` and skip that step if so?

Yeah it's not too bad. Though if one is trying to write array agnostic code, one would need to detect array type and then branch based on that. Still not too bad, but it is some extra boiler plate. This might get trickier the more array types that need handling.

jakirkham:
It would be good to get another review if people have time. Have pushed a few changes cleaning up stuff based on feedback, adding a few missing sections, etc.

jakirkham:
Also cc-ing @sofroniewn @tlambert03 @jni (in case this is of interest 😉)

jakirkham:
cc-ing @grlee77 as well

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,446 @@
...
+
+Summary
+-------
+
+[Last year](https://blog.dask.org/2019/08/09/image-itk) we experimented with
+Dask/ITK/Scikit-Image to perform large scale image analysis on a stack of 3D
+images.  Specifically, we looked at `deconvolution`, a common method to
+_deblur_ images.  Now, a year later, we return to these experiments with a
+better understanding of how Dask and CuPy can interact, enhanced serialization
+methods, and support from the open-source community.
```
stsievert: Maybe add a list of takeaways from this post?

> ... Now, a year later, we return to these experiments with a better understanding of how Dask and CuPy can interact, enhanced serialization methods, and support from the open-source community. Specifically, this post will mention the following:
>
> 1. ...
> 2. ...

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,446 @@
...
+theme: twitter
+---
+{% include JB/setup %}
+
+Summary
+-------
+
+[Last year](https://blog.dask.org/2019/08/09/image-itk) we experimented with
+Dask/ITK/Scikit-Image to perform large scale image analysis on a stack of 3D
+images.  Specifically, we looked at `deconvolution`, a common method to
```
stsievert: Style nit and suggested diff (no monospace styling):

``` diff
- `deconvolution`
+ deconvolution
```

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,446 @@
...
+
+```python
+def deconvolve(img, psf=None, iterations=20):
+    """
+    """
+    # 2D vs 3D psfs
+    if len(psf.shape) == 2:
+        psf = psf
+    else:
+        psf = psf[0]
```
stsievert: Style nit and suggested diff:

``` diff
-     # 2D vs 3D psfs
-     if len(psf.shape) == 2:
-         psf = psf
-     else:
-         psf = psf[0]
+ psf = psf if psf.ndim == 2 else psf[0]  # 2D vs 3D PSFs
```

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,446 @@
...
+    psf = np.pad(psf, tuple(zip(pad_l, pad_r)), 'constant', constant_values=0)
+
+    # padding centers the PSF so we no longer need to roll
+    # Center PSF at pixel (0,0)
+    for i in range(psf.ndim):
+        psf = np.roll(psf, psf.shape[i]//2, axis=i)
+
+    # Convolution requires FFT of the PSF
+    psf = np.fft.rfftn(psf)
+    psf = psf[None]
```
stsievert: What does this indexing do? I'm not familiar with the details of `rfftn`. A comment might help here.

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,446 @@
...
+    return img_decon
+```
+
+For a 1.3 GB image we measured the following:
+
+- CuPy ~3 seconds for 20 iterations
+- NumPy ~36 seconds for 2 iterations
+
+We see 10x increase in speed for 10 times the number of iterations -- very
+close to our desired 100X speedup!  Let's explore how this implementation
```
stsievert: Style nit: the "10x" and "100X" have different multiplication signs. I suggest using a consistent style (I use the UTF multiplication sign, "×" (HTML: `&#215;`/`&times;`, Unicode: `U+00D7`)).

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,446 @@
...
+```
+out = da.map_blocks(deconvolve,
+                    c_imgs,
+                    c_psf,
+                    iterations=20,
+                    dtype=cp.float32)
+```
+
+With Dask and multiple GPUs, we measured deconvolution of an 8GB image in ~10
+seconds! But this is just the first step towards accelerated image science.
```
stsievert: What are the next steps? Suggestion: I might move this below the image and expand there.

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,446 @@
...
+
+<a href="/images/deconvolve.png">
+    <img src="/images/deconvolve.png" width="100%"></a>
+
+IMAGE DESCRIPTION
```
stsievert: I am especially interested to see this! A higher resolution image might help too.

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,446 @@
...
+(RL)](https://en.wikipedia.org/wiki/Richardson%E2%80%93Lucy_deconvolution)
+deconvolution algorithm from ITK and
+[Scikit-Image](https://github.com/scikit-image/scikit-image/blob/master/skimage/restoration/deconvolution.py#L329).
+We left off at theorizing how GPUs could potentially help accelerate these
+workflows.  Starting with Scikit-Image's implementation, we naively tried
+replacing `scipy.signal.convolve` calls with `cupyx.scipy.ndimage.convolve`,
+and while performance improved, it did not improve _significantly_ -- that is,
+we did not get the 100X speed we were looking for.
+
+As it often turns out in mathematics a problem that can be inefficient to solve
```
GenevieveBuckley: These last two introduction paragraphs are a lot clearer now 😄

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,446 @@
...
+        psf = psf[0]
+
+    # pad psf with zeros to match image shape
+    # use NumPy for padding
+    pad_l, pad_r = np.divmod(np.array(img[0].shape) - np.array(psf.shape), 2)
+    pad_r += pad_l
+
+    psf = np.pad(psf, tuple(zip(pad_l, pad_r)), 'constant', constant_values=0)
+
+    # padding centers the PSF so we no longer need to roll
```
GenevieveBuckley: This comment says "we no longer need to roll" the psf, but three lines later there is a roll applied. This seems contradictory?

VolkerH:
Hi, 
just saw this and taking the liberty to comment as it is something I am interested in. 
I have a few questions/comments:

* With `map_blocks` there will be artefacts at block boundaries. I would suggest mentioning `map_overlap` in addition to `map_block`, see [here](https://github.com/hammerlab/flowdec/blob/master/python/examples/notebooks/Tile-by-tile%20deconvolution%20using%20dask.ipynb)
* I've been experimenting with cupy decon code and saw a significant performance increase after this FFT plan caching implementation https://github.com/cupy/cupy/pull/3730 was merged into cupy recently. I am not sure whether your benchmarks reflect this already.
* When reading the `deconvolve` implementation I stumbled over all the `np.`-code and wondered how this is going to work with `cupy`. I guess the answer was given a few lines above with the link to  `NEP-18` but I think it would be helpful to elaborate a bit more on that and translate it for beginner/intermediate numpy/cupy users who may not know  the details of the underlying dispatch mechnisms. The `NEP` is not easy reading and one needs to follow a link away from the blog post.

_posts/2020-10-23-deconvolution.md
```diff
@@ -0,0 +1,446 @@
...
+Dask over multiple GPUs.
+
+Help from Open-Source
+---------------------
+
+An FFT RL equivalent has been around for some time and the good folks at the
+[Solar Dynamics Observatory](https://sdo.gsfc.nasa.gov/mission/instruments.php)
+built and shared a NumPy/CuPy implementation as part the [Atmospheric Imaging
+Assembly](https://aiapy.readthedocs.io/en/v0.2.0/_modules/aiapy/psf/deconvolve.html)
+Python package (aiapy).  We slightly modified their implementation to handle 3D
```
VolkerH: I cannot follow this. You mention that you support 3-dimensional point spread functions in the text, but in the deconvolutoin code you just select one slice if the PSF has dimensions other than 2? And then there is padding in x and y dimensions. So are you trying to do 2d deconvolution on individual slices of the volume? In particular for the 3D microscopy dataset further down I would assume that the PSF is indeed 3 dimensional?


Edit: somehow I had an issue highlighting the correct region this comment was for. It was supposed to include both the text that mentions supporting 3D PSFs as well as the code.

_posts/2020-10-23-deconvolution.md
jakirkham:
```suggestion
images.  Specifically, we looked at deconvolution, a common method to
```

jakirkham:
> * With `map_blocks` there will be artefacts at block boundaries. I would suggest mentioning `map_overlap` in addition to `map_block`, see [here](https://github.com/hammerlab/flowdec/blob/master/python/examples/notebooks/Tile-by-tile%20deconvolution%20using%20dask.ipynb)

In this case we were able to pick chunks that were large enough in the dimensions where we were applying the PSF. So avoided an overlap on them.

> * I've been experimenting with cupy decon code and saw a significant performance increase after this FFT plan caching implementation [cupy/cupy#3730](https://github.com/cupy/cupy/pull/3730) was merged into cupy recently. I am not sure whether your benchmarks reflect this already.

Nope we haven't. That definitely seems useful though. It's not in any release yet. Though we could make a brief note about this.

> * When reading the `deconvolve` implementation I stumbled over all the `np.`-code and wondered how this is going to work with `cupy`. I guess the answer was given a few lines above with the link to  `NEP-18` but I think it would be helpful to elaborate a bit more on that and translate it for beginner/intermediate numpy/cupy users who may not know  the details of the underlying dispatch mechnisms. The `NEP` is not easy reading and one needs to follow a link away from the blog post.

Yeah we are relying on NEP-18. There's [a separate blogpost about the NEP]( https://blog.dask.org/2019/03/18/dask-nep18 ), which we could reference as well if that would help.

_posts/2020-10-23-deconvolution.md
jakirkham: Yeah the code probably needs to be updated. Initially we had a PSF that had an extra singleton dimension for some reason that we removed. We did some cleanup of the code, but we may have missed some things still. Thanks for pointing that out!

_posts/2020-10-23-deconvolution.md
jakirkham: Good point! Will take another look

_posts/2020-10-23-deconvolution.md
jakirkham:
```suggestion
close to our desired 100x speedup!  Let's explore how this implementation
```

_posts/2020-10-23-deconvolution.md
jakirkham: I find it slightly easier in the blog to see this spelled out as opposed to one line. That said, there may be other changes here anyways.

_posts/2020-10-23-deconvolution.md
jakirkham: This adds a singleton dimension. That said, I'm not sure we still need this. Will recheck. Thanks!

VolkerH:
> > * With `map_blocks` there will be artefacts at block boundaries. I would suggest mentioning `map_overlap` in addition to `map_block`, see [here](https://github.com/hammerlab/flowdec/blob/master/python/examples/notebooks/Tile-by-tile%20deconvolution%20using%20dask.ipynb)
> 
> In this case we were able to pick chunks that were large enough in the dimensions where we were applying the PSF. So avoided an overlap on them.

Ok, I missed this as I only discovered that you are only deconvolving single slices after that comment.

> > * I've been experimenting with cupy decon code and saw a significant performance increase after this FFT plan caching implementation [cupy/cupy#3730](https://github.com/cupy/cupy/pull/3730) was merged into cupy recently. I am not sure whether your benchmarks reflect this already.
> 
> Nope we haven't. That definitely seems useful though. It's not in any release yet. Though we could make a brief note about this.

This is already in the cupy 8.0.0 release which has been out for a few weeks! Simply upgrading should speed things up.

> Yeah we are relying on NEP-18. There's [a separate blogpost about the NEP](https://blog.dask.org/2019/03/18/dask-nep18), which we could reference as well if that would help.

I think this would help. But I would also suggest just adding an additional one or two sentences to emphasize a bit more how NEP-18 is the magic sauce that allows the `np.` code to operate on `cp.`-arrays by dispatching cupy functions. It would have helped me to have this additional context without following a link on my first read through the text.

BTW, I should say thank you as your WIP blog post made me aware of this possibility. I had previously spent  some effort converting some pieces of code to  `cupy` by importing either the cupy or the numpy functions. Your post showed me that this was completely unnecessary. 

I don't know what your overall goal is, i.e. whether it is showing the general NEP-18 dispatching mechanism and how it integrates with dask or to provide a real-world example for large-scale deconvolution. If it is the latter, I think an example with a 3D PSF and overlap of the tiles would be useful. However, if it is the former, those additional details may obscure your main point.

_posts/2020-10-23-deconvolution.md
grlee77: I think you should be able to remove the calls to `np.pad` and `np.roll` and just directly call rfftn with the `s` (shape) argument in rfftn:

```Python
psf = np.fft.rfftn(psf, s=img[0].shape)
```
In general indexing with `img[0]` and then adding a dimension back with `psf[None]` seems to assume a 2D PDF

grlee77:
@quasiben, do you have a link where I can download the same test data used in the blog post? I would like to download it locally to try out some things here.

grlee77:
Feel free to use the following variant where the axes of `img` that correspond to the PSF can be specified explicitly via an `axes` argument. If not provided, it is assumed that the psf corresponds to the last `psf.ndim` indices in `img`. This version should just work for e.g. `2d img/2d psf` or `3d img/2d psf` or `3d img/3d psf`, etc. If the 3d image is actually a stack of 2d images, but the stack is not along the first axis then the user would need to specify `axes`.

I understand if checking for valid axes using `normalize_axis_index`, etc. may be a bit more complicated than desired for the purposes of the blog post, though.

```Python
from numpy.core.multiarray import normalize_axis_index


def deconvolve(img, psf=None, iterations=20, axes=None):
    """
    """
    if psf.ndim > img.ndim:
        raise ValueError("psf cannot have more dimensions than img.")

    if axes is None:
        # assume that the last psf.ndim axes correspond to the PSF
        axes = tuple(range(img.ndim - psf.ndim, img.ndim))
    else:
        # use user-specified axes for the psf
        axes = tuple(normalize_axis_index(ax, img.ndim) for ax in axes)
        axes = sorted(axes)
        if psf.ndim != len(axes):
            raise ValueError(
                "The number of provided axes should match psf.ndim"
            )

    # add singleton dimensions along any additional axes
    psf_shape = [1,] * img.ndim
    for n, ax in enumerate(axes):
        psf_shape[ax] = psf.shape[n]
    psf = psf.reshape(psf_shape)

    # Convolution requires FFT of the PSF
    fft_shape = tuple(img.shape[ax] for ax in axes)
    psf = np.fft.rfftn(psf, s=fft_shape, axes=axes)
    psf_conj = psf.conj()

    img_decon = np.copy(img)
    for _ in range(iterations):
        ratio = (
            img / np.fft.irfftn(np.fft.rfftn(img_decon, axes=axes) * psf,
                                axes=axes)
        )
        img_decon *= np.fft.irfftn(np.fft.rfftn(ratio, axes=axes) * psf_conj,
                                   axes=axes)
    return img_decon
```

_posts/2020-10-23-deconvolution.md
GenevieveBuckley: Ooooh, random extra singleton dimensions make a lot more sense. Definitely be good to clean that up before this goes out!

quasiben:
@grlee77 apologies for the delay.  I got approval from Hari's Lab and you can download from s3:

> s3://dask-microscopy/y1z1_C1_A.zarr

Full list of files:
```python
import s3fs
fs = s3fs.S3FileSystem(anon=True)
fs.ls('s3://dask-microscopy')
['dask-microscopy/AVG_PSF.tif',
 'dask-microscopy/PSF.tif',
 'dask-microscopy/y1z1_C1_A.zarr']
```

grlee77:
Thanks @quasiben

You may also be interested in a non-iterative FFT-based deconvolution technique that was recently submitted to scikit-image by @alexjarosch (scikit-image/scikit-image#5026).

Yesterday, I made a PR against that branch that adds 3D support (https://github.com/alexjarosch/scikit-image/pull/1) (currently it requires `img.ndim == psf.ndim`, so please add a singleton dimension to the PSF if the image is 3D, but PSF is 3D). I haven't tried it on the GPU yet, but suspect it is probably an order of magnitude faster than Richardson-Lucy just by not having to perform several iterations.

jakirkham:
> Yesterday, I made a PR against that branch that adds 3D support ([alexjarosch/scikit-image#1](https://github.com/alexjarosch/scikit-image/pull/1)) (currently it requires `img.ndim == psf.ndim`, so please add a singleton dimension to the PSF if the image is 3D, but PSF is 3D). I haven't tried it on the GPU yet, but suspect it is probably an order of magnitude faster than Richardson-Lucy just by not having to perform several iterations.

Interesting, probably need to go read this paper then 😄

grlee77:
> Interesting, probably need to go read this paper then 

The paper also covers other penalties that do require either conjugate-gradient (CG) or iteratively-reweighted least squares (IRLS) to solve. This specific 2-norm penalty version is non-iterative, though.

leofang:
> > > * I've been experimenting with cupy decon code and saw a significant performance increase after this FFT plan caching implementation [cupy/cupy#3730](https://github.com/cupy/cupy/pull/3730) was merged into cupy recently. I am not sure whether your benchmarks reflect this already.
> > 
> > 
> > Nope we haven't. That definitely seems useful though. It's not in any release yet. Though we could make a brief note about this.
> 
> This is already in the cupy 8.0.0 release which has been out for a few weeks! Simply upgrading should speed things up.

Hi @VolkerH Thanks for bringing this up and reporting the performance for the cache I added! Glad to learn it helps. You are right that it's merged to CuPy master but was also backported to v8.0. Do you have a minimum code for me to benchmark, and what's the speed up that you observed?

VolkerH:
> Hi @VolkerH Thanks for bringing this up and reporting the performance for the cache I added! Glad to learn it helps. You are right that it's merged to CuPy master but was also backported to v8.0. Do you have a minimum code for me to benchmark, and what's the speed up that you observed?

Hi,
it is not really a "minimum" code sample, but the two files that constitute [this PR](https://github.com/david-hoffman/pyDecon/pull/6) are self-contained. The notebook downloads sample data and runs `%timeit`. The current version of the notebook already reflects the cached FFT plan. 
Regarding the exact speed-up: I destroyed my notes on that by accident when I deleteted a forked repo where I kept this. Will try to run it again.

edited to clarify ambiguous wording.

_posts/2020-10-23-deconvolution.md
jakirkham: Ok I think we've addressed this issue. Wound up doing a bit of rewriting of this function. The resulting code is a bit simpler and avoids branches around dimensions. Hopefully this makes more sense. Please let us know if anything else sticks out here 🙂

_posts/2020-10-23-deconvolution.md
jakirkham: Dropped this line and explained rolling. Hopefully that's clearer :)

_posts/2020-10-23-deconvolution.md
jakirkham: What do you think about leaving them in from an education standpoint? It seems useful for people to see the steps explicitly. Maybe we can make a note about using `s=`?

Yeah this code has been removed in a more recent rewrite. Please let us know if anything else sticks out :)

_posts/2020-10-23-deconvolution.md
jakirkham: This code has since been dropped

_posts/2020-10-23-deconvolution.md
jakirkham: Does `np.fft.rfftn(psf, s=img.shape)` also recenter the PSF around the origin? Or does one still need to multiply by a phase offset in FFT space to fix that? Just using `s` seems to move the deconvolved image somehow.

_posts/2020-10-23-deconvolution.md
jakirkham: Added a description. Also having included a new image. Not sure whether it is better. Please let us know :)

_posts/2020-10-23-deconvolution.md
jakirkham: It really depends on the problem being solved. One possibility might be to do segmentation followed by operations on the resulting label image.

jakirkham: Moved text below image

_posts/2020-10-23-deconvolution.md
jakirkham: Good idea. Added some takeaways below :)

quasiben:
In the last handful of commits @jakirkham implemented a 3D deconvolution using (and not flattening) a 3D PSF.  Ideally, we could release this tomorrow unless there is more feedback

_posts/2020-11-12-deconvolution.md
```diff
@@ -229,9 +229,11 @@ imgs = da.from_zarr("/public/NVMICROSCOPY/y1z1_C1_A.zarr/")
 
 From the Dask output above you can see the data is a z-stack of 950 images
 where each slice is 2048x2048.  For this data set, we can improve GPU
-performance if we operate on larger chunks.  As we did our work on a DGX2,
-which has 16 GPUs, we can comfortably fit the data and perform deconvolution on
-each GPU if we `rechunk` the data accordingly:
+performance if we operate on larger chunks.  Additionally, we need to ensure
+the chunk are are least as bif as the PSF which in this case is,
```
VolkerH: bif -> big

quasiben: Thanks @VolkerH!

_posts/2020-11-12-deconvolution.md
```diff
@@ -0,0 +1,435 @@
...
+replacing `scipy.signal.convolve` calls with `cupyx.scipy.ndimage.convolve`,
+and while performance improved, it did not improve _significantly_ -- that is,
+we did not get the 100X speed we were looking for.
+
+As it often turns out in mathematics a problem that can be inefficient to solve
+in one representation can often be made more efficent by transforming the data
+beforehand. In this new representation we can solve the same problem
+(convolution in this case) more easily before transforming the result back into
+a more familiar representation. When it comes to convolution, the
+transformation we apply is called [Fast-Fourirer Transform
```
VolkerH: Fourirer should be Fourier

_posts/2020-11-12-deconvolution.md
```diff
@@ -0,0 +1,435 @@
...
+(FFT)](https://en.wikipedia.org/wiki/Fast_Fourier_transform). Once this is
+applied we are able to convolve data using a simple multiplication.
+
+As it turns out this FFT transformation is extremely fast on both CPUs and
+GPUs. Similarly the algorithm we can write with FFTs are accelerated.  Despite
```
VolkerH: The algorithm ... *is* accelerated.

Note though that to people "trained in the art" this whole paragraph may look a bit odd, as it is standard practice to do deconvolution in FFT space.

jakirkham: Yeah that's fair. Would you recommend dropping it then?

jakirkham: One data point worth looking at here is SciPy actually chooses between doing it in real or FFT space ( https://github.com/scipy/scipy/pull/5608 ), which is [what scikit-image calls to do `richardson_lucy`]( https://github.com/scikit-image/scikit-image/blob/ed889ba3deb17cd94d1f0f9b730f6564585936ee/skimage/restoration/deconvolution.py#L379 ). So there may still be cases where real is more efficient.

_posts/2020-11-12-deconvolution.md
jakirkham: Thanks! Fixed :)

_posts/2020-11-12-deconvolution.md
VolkerH: > Yeah that's fair. Would you recommend dropping it then?

I would suggest rephrasing it. Currently it reads a bit as "we discovered that one can actually do this in Fourier space and it turns out to be much faster"  to paraphrase it very loosely. I would simply phrase it more as "In contrast, to our previous example, we perform deconvolution in fourier space, a commonly used approach ...." and then keep some of the explanations why it is fast as they are still useful for readers who may not be familiar with the concept that a convolution of two signals in one domain is equivalent to multiplication of those signals in the Fourier-transformed domain. So in essence, I  would  avoid any phrasing that may create the impression that there is something novel about this approach.

jakirkham: Ok I tried to add a line indicating this is a commonly used technique. Though this is also alluded to in the final sentence of the paragraph as well.

VolkerH: Yes, I like that better.

_posts/2020-11-12-deconvolution.md
```diff
@@ -0,0 +1,436 @@
...
+</td>
+
+Next, we convert to `float32` as the data may not already be of floating point
+type. Also 32-bit is a bit faster than 64-bit when computing and saves a bit on
+memory.  Below we map `cupy.asarray` onto each block of data.  `cupy.asarray`
+moves the data from host memory (NumPy) to the device/GPU (CuPy).
+
+```python
+imgs = imgs.astype(np.float32)
+c_imgs = imgs.map_blocks(cupy.asarray)
```
grlee77:
```suggestion
c_imgs = imgs.map_blocks(cp.asarray)
```
cupy was imported as `cp` above

_posts/2020-11-12-deconvolution.md
```diff
@@ -0,0 +1,436 @@
...
+What we now have is a Dask array composed of 16 CuPy blocks of data.  Notice
+how Dask provides nice typing information in the SVG output.  When we moved
+from NumPy to CuPy, the block diagram above displays `Type: cupy.ndarray` --
+this is a nice sanity check.
+
+The last piece we need before running the deconvolution is the PSF which should
+also be loaded onto the GPU:
+
+```python
+psf = skimage.io.imread("/public/NVMICROSCOPY/PSF.tif")
```
grlee77, line 361:
```suggestion
import skimage.io

psf = skimage.io.imread("/public/NVMICROSCOPY/PSF.tif")
```

_posts/2020-11-12-deconvolution.md
```diff
@@ -0,0 +1,436 @@
...
+
+By adding a `map_blocks` call to our array, we can move our data _back_ from
+GPU to CPU (device to host).
+
+```python
+def cupy_to_numpy(x):
+    import cupy as cp
+    return cp.asnumpy(x)
+
+np_out = out.map_blocks(cupy_to_numpy, meta=out)
```
grlee77, line 409:
```suggestion
np_out = out.map_blocks(cp.asnumpy, meta=out)
```

why not just pass cp.asnumpy directly to map_blocks?

_posts/2020-11-12-deconvolution.md
grlee77, line 361:
```suggestion
import skimage.io

psf = skimage.io.imread("/public/NVMICROSCOPY/PSF.tif")
```

_posts/2020-11-12-deconvolution.md
grlee77:
```suggestion
c_imgs = imgs.map_blocks(cp.asarray)
```
cupy was import as `cp` above

_posts/2020-11-12-deconvolution.md
jakirkham: Thanks for taking a look :)

_posts/2020-11-12-deconvolution.md
```diff
@@ -0,0 +1,436 @@
...
+out
+```
+
+<a href="/images/deconvolve.png">
+    <img src="/images/deconvolve.png" width="100%"></a>
+
+The image above is taken from a mouse intestine.
+
+With Dask and multiple GPUs, we measured deconvolution of an 8GB image in ~10
+seconds! But this is just the first step towards accelerated image science.
```
grlee77: It would be nice to have a reference for CPU performance to compare, even if it is just for 1 or 2 iterations. Not sure what the CPU configuration is in the DGX2?

grlee77 (commented):
The post looks good. I left a couple of minor fixes needed for the code to work if run sequentially as presented.

grlee77:
I do have some question of whether the 3D PSF is helping, though. Here is what I get plotting the slice at imgs[450, 512:-512, 512:-512] for the 3D PSF vs. using `psf=skimage.io.imread('AVG_PSF.tif')[np.newaxis, ...]`

I am not a microscopist so I don't really know what I am looking at here, but the middle image produced by the 2D PSF looks substantially sharper to me:
deconv_compare

However, in an orthogonal view some things definitely look sharper for the 3D psf case (not shown). It doesn't matter as far as the computational point being made in the blog post, but it would be good to get some feedback from a microscopist on whether these results look as expected.

_posts/2020-11-12-deconvolution.md
```diff
@@ -0,0 +1,436 @@
...
+we did not get the 100X speed we were looking for.
+
+As it often turns out in mathematics a problem that can be inefficient to solve
+in one representation can often be made more efficent by transforming the data
+beforehand. In this new representation we can solve the same problem
+(convolution in this case) more easily before transforming the result back into
+a more familiar representation. When it comes to convolution, the
+transformation we apply is called [Fast-Fourier Transform
+(FFT)](https://en.wikipedia.org/wiki/Fast_Fourier_transform). Once this is
+applied we are able to convolve data using a simple multiplication.
```
VolkerH, line 45: Apologies if this comes across as pedantic, but I just re-read this after looking at a few hours: there are still some phrases that I consider slightly unfortunate. You mention at several points that one can speed up convolutions by going to the Fourier domain (I just picked one example here, that statement is present multiple times with slight variation). 
Convolutions are a mathematical operation typically reprenset by nested integrals (or sums in the discrete case). We are not really speeding up convolutions by going to the Fourier domain, we are avoiding them altogether. The result (when transformed back into the spatial domain) is (boundary effects and other numerical issues aside) the same.
So what is sped up are signal filtering operations, through *not* performing convolutions.
The algorithm in its totality should still be called "deconvolution" though as it is not referring to the specific mathematical operation, even if implemented without explicit convolutions.

VolkerH:
> 
> 
> I do have some question of whether the 3D PSF is helping, though. Here is what I get plotting the slice at imgs[450, 512:-512, 512:-512] for the 3D PSF vs. using `psf=skimage.io.imread('AVG_PSF.tif')[np.newaxis, ...]`
> 
> I am not a microscopist so I don't really know what I am looking at here, but the middle image produced by the 2D PSF looks substantially sharper to me:
> deconv_compare
> 
> However, in an orthogonal view some things definitely look sharper for the 3D psf case (not shown). It doesn't matter as far as the computational point being made in the blog post, but it would be good to get some feedback from a microscopist on whether these results look as expected.

I agree that in those screenshots the middle image looks more crisp.
Without access to the data, it is difficult to answer.

Many things can go wrong, maybe the 3D PSF isn't really that good. The blog article doesn't seem to mention exactly what type of microscope this is from (lightsheet, confocal, spinning disk) and how the PSF was obtained.

Also, it is quite possible that more RL iterations might be needed with the 3D PSF to get a comparable result. There is often no real convergence, there's usually subjective eyeballing involved how many iterations are necessary for a given dataset.

Most microscopy data has much less resolution in the Z direction, so showing a slice across the XZ or YZ in addition to XY might show that the structures look less blurred in depth.  In a single XY slice, this may look like some faint cells are disappearing as they are only out-of-focus light from an adjacent stack.

grlee77:
Here is an orthogonal view (slice [:, 1024, 512:-512]). I have not looked into varying iterations. So far for this image and the prior one, it was 20 iterations as in the post.

deconv_compare_ortho

_posts/2020-11-12-deconvolution.md
VolkerH: However, I might be too sensitive here, the wikipedia article https://en.wikipedia.org/wiki/Convolution_theorem uses similar phrasing to yours: 
```
This formulation is especially useful for implementing a numerical convolution on a computer: The standard convolution algorithm has quadratic computational complexity. With the help of the convolution theorem and the fast Fourier transform, the complexity of the convolution can be reduced from O ( n 2 ) {\displaystyle O\left(n^{2}\right)} {\displaystyle O\left(n^{2}\right)} to O ( n log ⁡ n ) {\displaystyle O\left(n\log n\right)} {\displaystyle O\left(n\log n\right)},
```
So maybe disregard my overly pedantic interpreation above.

_posts/2020-11-12-deconvolution.md
```diff
@@ -0,0 +1,436 @@
...
+    depth=tuple(np.array(c_psf.shape) // 2),
+    boundary="periodic"
+)
+out
+```
+
+<a href="/images/deconvolve.png">
+    <img src="/images/deconvolve.png" width="100%"></a>
+
+The image above is taken from a mouse intestine.
```
stsievert, line 384: Could you add a little more context? Why is the mouse intestine important and why does the scientist want a deblurred image?

_posts/2020-11-12-deconvolution.md
jakirkham: I don't know that I would agree we are avoiding convolutions. Multiplying a kernel by an image in Fourier space is the same as performing the nested integral in real space. Both constitute a convolution. How we turn the crank to get that result may differ, but the result is still a convolution of the two.

_posts/2020-11-12-deconvolution.md
jakirkham: @quasiben, do you have thoughts here?

_posts/2020-11-12-deconvolution.md
jakirkham: Here's the [specs for a DGX-2]( https://www.nvidia.com/content/dam/en-zz/Solutions/Data-Center/dgx-2/dgx-2-print-datasheet-738070-nvidia-a4-web-uk.pdf ).

Basically we have 16 Tesla V100s and 24 Dual Intel Xeon Platinum 8168. I guess this raises a question of how we compare these. Do we just use all GPUs and compare to all CPUs? Should we use an equal number of both? Maybe we should do something different?

cc @quasiben (in case you have thoughts here as well)

jakirkham:
The latest image we pushed used 100 iterations. That seemed to help some more. Though you may be right that a better PSF could be needed.

Anyways I don't want us to get to bogged down on the analysis of this data. The goal is more to present how one might approach this problem and leverage the Python ecosystem and recent NEPs to tackle this kind of problem. I think those of you reviewing here are more than capable of iterating on this and improving it further 😄

GenevieveBuckley:
> Anyways I don't want us to get to bogged down on the analysis of this data. The goal is more to present how one might approach this problem and leverage the Python ecosystem and recent NEPs to tackle this kind of problem. I think those of you reviewing here are more than capable of iterating on this and improving it further 😄

This is reasonable, I think you've given enough of the "bones" here for someone else to go on and try to use it in their own work.

_posts/2020-11-12-deconvolution.md
GenevieveBuckley: I'm not sure I'd agree with these suggestions. For me as a reader, I don't think it's necessary to add any more detail here.

I think it's more or less self-evident why scientists would want a deblurred (i.e. better quality) image. 

Why someone might be choosing to look at a mouse intestine probably depends a lot on what their research question is, which is probably getting off-topic to include in this post.

VolkerH:
> w one might approach this problem and leverage the Python ecosystem and recent NEPs to tackle this kind of problem

That is perfectly fine! It described splitting up the computation, using cupy via NEP (which I didn't know about), super useful.
As long as the deconvolved result shown doesn't look really bad or even worse than the original as that may raise eyebrows.

_posts/2020-11-12-deconvolution.md
VolkerH: Ok. I stand corrected.
Quoting wikipedia again:
*The term convolution refers to both the result function and to the process of computing it.*
I was too rigorous above in referring only the process as convolution.  
All good.

jakirkham: No worries. We appreciate the thoroughness 🙂

_posts/2020-11-12-deconvolution.md
jakirkham: FWIW I've asked my colleagues if they can shed some more light here.

That said, would agree with the points Genevieve has made above.

jakirkham:
Thanks everyone for your feedback! 😄

Have gone ahead and published it. If you spot anything else, please let us know 🙂

GenevieveBuckley:
Naive question: how does what's implemented here differ from something like [pycudadecon](https://github.com/tlambert03/pycudadecon)?

Would we recommend people use one instead of the other in certain circumstances?

@jakirkham @quasiben @tlambert03

tlambert03:
To my understanding, the primary difference is that what was done here scales to clusters of GPUs using dask maps blocks to handle large datasets.  pycudadecon is limited to a single GPU (which can easily run out of memory) ... but has a few additional conveniences around data and PSF/OTF preparation

tlambert03:
(also: pycudadecon wraps a [specific precompiled cuda/c++ library](https://github.com/scopetools/cudaDecon)... whereas this uses cupy and will almost certainly be easier to install/maintain)

jakirkham:
Agree with Talley

One might consider a modification to this workflow that use pycudadecon as the operation to map over blocks

GenevieveBuckley:
Thanks @tlambert03 , I hadn't realised the point you make about single vs multiple GPUs.

> One might consider a modification to this workflow that use pycudadecon as the operation to map over blocks

@jakirkham there is an [example here](https://napari.org/tutorials/applications/dask) that Talley wrote using pycudadecon & map_blocks. Not sure if you've seen this already, but if you scroll all the way to the bottom you'll see the full script.

jakirkham:
I hadn’t. That’s great! 😀

GenevieveBuckley:
@jakirkham you might also like to take a look at [pyDecon](https://github.com/david-hoffman/pyDecon) @VolkerH has been adding cupy support to it recently.

jakirkham:
Yeah I saw his PR ( https://github.com/david-hoffman/pyDecon/pull/6 ) 🙂

VolkerH:
It actually didn't need much support.
I originally rewrote it with a `from cupy import cp` instead of `from numpy import np`.
When I saw this blog post here I learned about NEP-18 and realised that much of that was not necessary. 

thanks for the mention ... just saw that this got merged

jakirkham:
Glad to hear the NEP-18 work has been useful (cc @pentschev who has done a lot of work on this 🙂)

Indeed saw that it got merged. Nice work! 😄

jakirkham:
Are there ways we could help make the NEP-18 work more visible so folks like yourself could pick it up more easily? Peter had written [a post about it]( https://blog.dask.org/2019/03/18/dask-nep18 ). How can we better draw attention to resources like this?

VolkerH:
Not sure how to tackle the disemmination of such new features.

For me the blog post did it (but I had to read it carefully during draft phase). But that was probably coincidence that I happened to see this.

Usually, when I find out about some python feature or some package I didn't know of before ... I often find out from multiple sources at once. I am not sure whether that is bias (only paying proper attention when something is mentioned several times) or whether it is due to a sudden collective awareness in the community that makes a particular feature go viral. E.g. I sometimes see a lot of people I follow on Github starring a repo at the same time. On some occassions I later realised that it had been mentioned on a podcast a few days earlier. 
Having said that, it proabably helps having some very applied and practical examples. Sharing information such as "with NEP-18 numpy can dispatch functions for other array types" is probably too dry and abstract to grab the attention of people if they are not specifically looking for that information, even people who may need just such a feature and have the background to understand the technicalities. 

Now I typed a lot of text that can be summarized as "I really don't know" :-)

GenevieveBuckley:
> Having said that, it proabably helps having some very applied and practical examples. Sharing information such as "with NEP-18 numpy can dispatch functions for other array types" is probably too dry and abstract to grab the attention of people

Yes, this

jakirkham:
Yep that's the line I was drawn to as well. What other applied examples would be useful? Could we come up with a short list of ideas?

I think one thing we will inevitably run into is the people doing the work on NEPs, implementations, etc. probably have limited cycles and may not be experts on the applied cases. So it may make sense to pair them up with people in the community to help fill out these posts. So could we identify subject matter experts from the community willing to help with this work?

tlambert03:
> > @tlambert03 said: (pycudadecon wraps a specific precompiled cuda/c++ library... whereas this uses cupy and will almost certainly be easier to install/maintain)

> @jakirkham said: One might consider a modification to this workflow that use pycudadecon as the operation to map over blocks

just an update here: [cudadecon](https://anaconda.org/conda-forge/cudadecon)/[pycudadecon](https://anaconda.org/conda-forge/pycudadecon) is now available on conda-forge ... so the maintainance burden has been dramatically reduced :)  Would be fun to pick this up again and try mapping `pycudadecon.decon` over blocks

PR #84, opened 2021-03-16
Update slack invite link in 2021-03-04-the-life-science-community.md

GenevieveBuckley:
Update the slack invite link so it is automatically issued, instead of Gen having to do it manually.


jrbourbeau (approved):
Woo, thanks @GenevieveBuckley!

PR #93, opened 2021-05-10
Skeleton analysis

GenevieveBuckley:
This blogpost focuses on the science use case, and the prototype distributed skeleton analysis workflow.

@jni and I might write a part 2 blogpost, describing adding dask support to skan.


GenevieveBuckley:
cc @jni, want to take a look at this? It feels like simultaneously there is both too much and not enough detail (I don't really explain what a skeleton graph is, but do jump straight into the nitty gritty problem parts of the code)

I might like to add a link to a gist, once it's tidied up.

GenevieveBuckley:
Merging this, now that Marcus & Andrew have had a look over it (conversation in a separate email thread).

I've also added a repository with an example notebook, so the full implementation details are available: https://github.com/GenevieveBuckley/distributed-skeleton-analysis/

PR #94, opened 2021-05-10
Adding Dask support to a third party library: A case study with skan

GenevieveBuckley:
@jni I've made us a skeleton outline (get it?) for a skan + dask blogpost, so we can fill it in as we go. 

You should have permissions to push to my branch (if I haven't set it up right, let me know)


GenevieveBuckley:
I had this ear-marked as a good blogpost idea, but now that Juan and I have worked on it I don't think there were as many general lessons that other people would also be interested in as I'd originally imagined.

PR #98, opened 2021-05-21
Recap of the life science workshop at the Dask Summit 2021

GenevieveBuckley:
Blogpost recapping the life science workshop at the Dask Summit. This is primarily intended as an overview for people who weren't able to attend this time round.

### DIFF
--- a/_posts/2021-05-24-life-science-summit-workshop.md
+++ b/_posts/2021-05-24-life-science-summit-workshop.md
@@ -1,0 +1,127 @@
+---
+layout: post
+title: Life sciences at the 2021 Dask Summit
+author: Genevieve Buckley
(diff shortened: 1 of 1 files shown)


GenevieveBuckley:
@jakirkham let me know if you feel like there's anything important that should also be added

jakirkham:
I think there were a couple more people on the last day that were not on the list here, but don't recall bios atm

jakirkham:
Also not sure if we want to summarize what was discussed

GenevieveBuckley:
Yeah, there were definitely more people attending than we have short bios for (I did try and note that). Perhaps we should have a quick sync up about what was discussed. It did get hard to keep track especially when people split out into breakout rooms.

PR #103, opened 2021-06-30
Blogpost on ragged output (with overlapping array chunks)

GenevieveBuckley:
I thought I'd write up some of the key takeaways from discussions we've been having about overlapping array chunks producing ragged outputs. 

There didn't appear to be one obviously preferred way to do this, so I'm hoping this post will make the preferred path a little bit clearer for people doing similar work. (Originally I had some other stuff mixed in to this post, but I think that muddied the message too much).

Related discussions:
* https://github.com/dask/dask/issues/7772
* https://github.com/dask/dask/issues/7589
* https://github.com/dask/dask/issues/7589

### DIFF
--- a/_posts/2021-07-02-ragged-output.md
+++ b/_posts/2021-07-02-ragged-output.md
@@ -1,0 +1,128 @@
+---
+layout: post
+title: Ragged output: how to handle awkward shaped results
+author: Genevieve Buckley
+theme: twitter
+---
+{% include JB/setup %}
+
+## Executive Summary
+
+This blogpost explains some of the difficulties associated with distributed computation and ragged or irregularly shaped outputs. We present a recommended method for using Dask in these circumstances.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


GenevieveBuckley:
Since this is a relatively uncontroversial summary of previous discussions, I'm going to go ahead and merge this.

jakirkham:
cc @jpivarski (in case this is of interest to you and/or your community)

jpivarski:
Thanks for pointing me to it! This seems to be a "technical raggedness," though—the functions that generate the data return different length outputs, but they are to be logically viewed as a concatenated array. (Looks like a good solution to that, too.)

The HEP community, and presumably others, often have to deal with data whose _meaning_ is ragged: the data collection includes 3 of these, 5 of those, 4 of something else, etc., and they should not be viewed as a concatenated collection because that would lose information about what it is one wants to model. So that's a different topic, and hopefully we'll be talking more about Dask-based solutions for that in the nearish future.

Thanks again for the heads-up!

PR #104, opened 2021-07-02
Fix title formatting in post, change colon to comma

GenevieveBuckley:
I suspect the reason the newest post title isn't being shown is that there is a colon `:` in the title. Testing this theory by changing it to a comma, I think that should fix it.

image


GenevieveBuckley:
Yep, that's done the trick. Working now!

image

PR #112, opened 2021-09-17
Fix broken mybinder links for dask-examples notebooks

GenevieveBuckley:
All the binder badge links are broken, this PR restores that functionality.

They broke for two reasons:
1. The dask-examples "main" branch was recently renamed (previously named "master"), and
2. A backwards incompatible change happened in the way binder generates/parses links ([see here for more details](https://discourse.jupyter.org/t/bug-path-to-a-notebook-file-functionality-unexepectedly-changed-not-working-for-notebooks-in-subfolders/10811))

Neither of these should be a problem going forward. However, we will need to be careful with the way people generate binder badges for dask-examples - if you use https://mybinder.org/ it will give you a link that only works for newer versions. We can either:
- Use the link formatting with `urlpath` like in this PR, or
- Update the pinned dependencies over at dask-examples to be more modern

Cross reference: https://github.com/dask/dask-examples/issues/196

### DIFF
--- a/_posts/2018-07-08-dask-dev.md
+++ b/_posts/2018-07-08-dask-dev.md
@@ -83,8 +83,8 @@ learning, etc.
 Now [Scott Sievert](https://stsievert.com/) is expanding the examples within
 the machine learning section.  He has submitted the following two so far:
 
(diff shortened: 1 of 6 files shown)


GenevieveBuckley:
(I'm not going to wait for discussion before merging, it's fixing a problem so we probably don't want to leave the links broken for longer than they need to be. I've checked the updated links individually & they all seem to be working now.)

PR #119, opened 2021-10-21
Blogpost on choosing good chunk sizes in Dask

### LINKED ISSUE #116: Blogpost idea: choosing good chunk sizes in Dask (turn this tweetorial into a blogpost)
Ian mentioned [this tweet](https://twitter.com/DataNerdery/status/1424953376043790341) to me today. I originally wrote it because I'd just given a tutorial, and lots of people were confused about how to choose good chunk sizes in Dask. Apparently that was helpful to a lot of people (this is supported by the twitter analytics stats, which are much higher than typical). 

For better discoverability and permanence, it might be good to have these tips in blogpost format (twitter is a bit ephemeral, and searches don't often unearth content there).

GenevieveBuckley:
Closes https://github.com/dask/dask-blog/issues/116

Lots of people, especially newcomers, are confused about how to choose good chunk/partition sizes in Dask. We should try and make information on this topic more well known.
TODO list (suggestions from reviewer comments):
- [x] Remove duplicate file, fix heading formatting
- [x] Add some code that shows how to see chunk sizes, how to change, it, etc.
- [x] Show pictures of the array HTML repr, visual representation of array chunks
- [x] Section on choosing the intial chunk size
    - [x] It's also super common for the data storage format to dictate chunk sizes already. Might be good to include a section on that, how it can be changed, whether rechunking is a valuable use of time, etc.
    - [x] More precise guidelines
    - [x] Rule of thumb calculation?
- [x] Talk about the cost of rechunking
- [x] Be more specific about when & why Dask spills data to disk
- [x] Images of the dask dashboard when things are not going well
- [x] "when this happens" -> "When we try to load too much data in memory on too few workers"
- [x] Add "OutOfMemory" error as a possible sign of problems


GenevieveBuckley:
This isn't a finished first draft, but it should give you enough idea of what topics are covered. If you think I should include extra detail, or something else, feedback is welcome.

GenevieveBuckley:
Draft is ready for review. I'll leave it open for comments for a few days, then merge it early next week.

jacobtomlinson (commented):
I really like this post, this comes up so often and it'll be great to have somewhere to point folks.

I see there is no code, just general high level advice. I think folks would benefit from some code that shows how to see chunk sizes, how to change, it, etc.

It's also super common for the data storage format to dictate chunk sizes already. Might be good to include a section on that, how it can be changed, whether rechunking is a valuable use of time, etc.

_posts/2021-10-25-choosing-chunk-and-partition-sizes.md
```diff
@@ -0,0 +1,79 @@
...
+If the foreman (the Dask scheduler) tells workers to go and fetch a single brick at a time, then bring each one to where the wall is being built, you can see how this is going to be very slow and inefficient! The workers are spending most of their time moving between the wall and the pile of bricks. Much less time is going towards doing the actual work of mortaring bricks onto the wall.
+
+Instead, we can do this in a smarter way. The foreman (Dask scheduler) can tell the workers to go and bring one full wheelbarrow load of bricks back each time. Now workers are spending much less time moving between the wall and the pile of bricks, and the wall will be finished much quicker.
+
+## Too big is also a problem
+
+If the Dask array chunks are too big, this is also bad. Why?
+Chunks that are too large are bad because then you are likely to run out of working memory.
+
+When this happens, Dask will try to spill data to disk instead of crashing.
```
guillaumeeb: Can this be **only** related to chunks being too large? Is this not because whatever the chunks size, you're trying to load too much data in memory on too few workers (or that Dask sometimes loads data too eagerly)?

For me too large chunks either throw OutofMemory errors, when a single chunk is too large to fit in memory, or either prevent Dask to effectively distribute the workload.

I agree there is a case, when the chunk size leads to chunks volume say about at half a worker memory, where you're likely to spill to disk each time a worker loads a chunk, or anyway too often, because of their size. But I'm not sure this is the principal problem you'll run into with too big chunks.

_posts/2021-10-25-choosing-chunk-and-partition-sizes.md
```diff
@@ -0,0 +1,79 @@
...
+
+If you're not very familiar with the Dask dashboard, or you just sometimes forget where to find certain dashboard plots (like the worker memory plot), then you'll probably enjoy these quick video tutorials:
+
+- [Intro to the Dask dashboard (18 minute video)](https://www.youtube.com/watch?v=N_GqzcuGLCY)
+- [Dask Jupyterlab extension (6 minute video)](https://www.youtube.com/watch?v=EX_voquHdk0)
+- [Dask dashboard documentation](http://distributed.dask.org/en/latest/diagnosing-performance.html)
+
+We recommend always having the dashboard up when you're working with Dask.
+It's a fantastic way to get a sense of what's working well, or poorly, so you can make adjustments.
+## Unmanaged memory
```
guillaumeeb:
```suggestion

## Unmanaged memory
```

guillaumeeb (commented):
Thanks for the post, really nice. 👍 

In addition to my inline comments, a few more:

- Could we provide some more precise guidelines, like for example: chunk size < 1MB is almost always bad, chunk size between 50MB and 1GB are generally good, chunk size over 2GB is almost always bad? I know this can vary a lot depending of the input data volume, you want handle 5GB of data as 5TB, and it also depends on the computation. But maybe some general rule of thumb? 
- We could even provide some calculation based on the input volume and number of workers/available worker memory? For example:
  - Chunks size should be lower than worker memory divided by 2 (or something like that),
  - You shouldn't end with more than 10,000 or 100,000 chunks, but aim for at least number of chunks equal to workers cores, and better workers cores times 2.
  - So a dataset of 5GB, 20 worker cores: 256MB max, 64MB sounds good, 
  - Dataset of 5GB and 200 worker cores: you probably have too much workers. But you could use 8MB or 16MB chunks.
  - Dataset of 5TB, 20 workers cores: too much data depending on workload, else try between 128MB and 2GB chunks. Also true for 200 cores.
  - I know these may not be true everytime, and really depends on the computation, but that could help some people?
- You've got two similar pages in you commit, not sure this is intended.
- Maybe put some more images:
  - one with the HTML output of a dask-array, which shows how that dataset is chunk, and the number of chunks, or event tasks that would be generated.
  - The Dask dashboard plots your mentioning, when it gets bad.

GenevieveBuckley:
> * Could we provide some more precise guidelines, like for example: chunk size < 1MB is almost always bad, chunk size between 50Mo and 1GB are generally good, chunk size over 2Go is almost always bad? I know this can vary a lot depending of the input data volume, you want handle 5GB of data as 5TB, and it also depends on the computation. But maybe some general rule of thumb?
> 
> * We could even provide some calculation based on the input volume and number of workers/available worker memory? For example:
>       
>   * Chunks size should be lower than worker memory divided by 2 (or something like that),
>   * You shouldn't end with more than 10,000 or 100,000 chunks, but aim for at least number of chunks equal to workers cores, and better workers cores times 2.
>   * So a dataset of 5GB, 20 worker cores: 256MB max, 64MB sounds good,
>   * Dataset of 5GB and 200 worker cores: you probably have too much workers. But you could use 8MBor 16MB chunks.
>   * Dataset of 5TB, 20 workers cores: too much data depending on workload, else try between 128MB and 2GB chunks. Also true for 200 cores.
>   * I know these may not be true everytime, and really depends on the computation, but that could help some people?

More specific rules of thumb would be great to include. Do other people generally agree with the rules of thumb Guillaume suggests here? @jacobtomlinson would you add anything extra?

_posts/2021-10-25-choosing-chunk-and-partition-sizes.md
GenevieveBuckley: I've tried to address this by:
1. Also mention that users may see out of memory errors if the chunk size is too large, and...
2. Clarifying that several causes can result in spilling to disk. I've tried to edit it to include your phrasing, so now it says *"When too much data is loaded in memory on too few workers, Dask will try to spill data to disk instead of crashing."*

GenevieveBuckley:
I've added a section with our very rough rules of thumb, and also content explaining that chunks should be aligned with the storage on disk. 

## Choosing an initial chunk size

Some very rough rules of thumb:

- When developing a new data processing pipeline, many people create a prototype using a small subset of the data, which may not involve Dask at all. This is helpful for quickly iterating on ideas, and also has the secondary benefit of giving you a clear idea of what size data can be processed easily for this workflow. You can use this knowledge to choose similar sized chunks in Dask.
- Some people have observed that chunk sizes below 1MB are almost always bad. Chunk size between 50MB and 1GB are generally good (and chunk sizes over 2GB may be too big - but this upper limit varies a huge amount depending on the computational resources you have available, so adjust for your circumstances)
- Upper bound: Avoid too very large task graphs. More than 10,000 or 100,000 chunks will perform poorly for this reason.
- Lower bound: To get the advantage of parallization, you need the number of chunks to at least equal the number of workers available (or better, the number of workers times 2). Otherwise, if the number of chunks is less than the number of workers, some workers will stay idle.
- The time taken to compute each task should be much larger than the time needed to schedule the task. The Dask scheduler takes roughly 1 millisecond to coordinate a single task, so a good task computation time would be measured in seconds (not milliseconds).

### Chunks should be aligned with array storage on disk
... (more content, see the full file if you want to keep reading)

GenevieveBuckley:
I think now it's a bit wordy and could benefit from good editing to make it shorter. Suggestions appreciated! We'll see how that goes, it's probably more important to get it out there sooner rather than later. Planning on publishing it this week.

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+## What are Dask array chunks?
+
+Dask arrays are big structures, made out of many small chunks.
+Typically, each small chunk is an individual [numpy array](https://numpy.org/), and they are arranged together to make a much larger Dask array.
+
+<img src="https://raw.githubusercontent.com/dask/dask/ac01ddc9074365e40d888f80f5bcd955ba01e872/docs/source/images/dask-array-black-text.svg" alt="Diagram: Dask array chunks" width="400" height="300" />
+
+You can find more information about Dask array chunks on this page of the documentation: https://docs.dask.org/en/latest/array-chunks.html
+
+### How do I nkow what chunks my array has?
```
guillaumeeb:
```suggestion
### How do I know what chunks my array has?
```

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Summary
+
+Confused about choosing [a good chunk size](https://docs.dask.org/en/latest/array-best-practices.html#select-a-good-chunk-size) for Dask arrays?
+
+Array chunks can't be too big (we'll run out of memory), or too small (the overhead introduced by Dask becomes overwhelming). So how can we get it right?
+
+It's a two step process:
+
+1. First, start by choosing a chunk size similar to data you know can be processed entirely within memory (i.e. without dask)
```
guillaumeeb:
```suggestion
1. First, start by choosing a chunk size similar to data you know can be processed entirely within memory (i.e. without Dask)
```

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+2. Then, watch the dask dashboard task stream and worker memory plots, and adjust if needed.
```
guillaumeeb:
```suggestion
2. Then, watch the Dask dashboard task stream and worker memory plots, and adjust if needed.
```

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+
+`arr.chunksize` shows the largest chunk size. For arrays where you expect roughly uniform chunk sizes, this is a good way to summarize chunk size information.
+
+`arr.chunks` shows fully explicit sizes of all chunks along all dimensions within the Dask array (see [item 3 here](https://docs.dask.org/en/stable/array-chunks.html#specifying-chunk-shapes)). This is more verbose, and is a good choice with arrays that have irregular chunks.
+
+## Choosing an initial chunk size
+
+Some very rough rules of thumb:
+
+- When developing a new data processing pipeline, many people create a prototype using a small subset of the data, which may not involve Dask at all. This is helpful for quickly iterating on ideas, and also has the secondary benefit of giving you a clear idea of what size data can be processed easily for this workflow. You can use this knowledge to choose similar sized chunks in Dask.
```
guillaumeeb:
```suggestion
- If you already created a prototype, which may not involve Dask at all, using a small subset of the data you intend to process, you'll have a clear idea of what size of data can be processed easily for this workflow. You can use this knowledge to choose similar sized chunks in Dask.
```

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+- Some people have observed that chunk sizes below 1MB are almost always bad. Chunk size between 50MB and 1GB are generally good (and chunk sizes over 2GB may be too big - but this upper limit varies a huge amount depending on the computational resources you have available, so adjust for your circumstances)
```
guillaumeeb: The link given at the beginning of the article even say "it is rare to see chunk sizes below 100 MB". Not sure how to bring this here, or if we should...

guillaumeeb:
```suggestion
- Some people have observed that chunk sizes below 1MB are almost always bad. Chunk size between 50MB and 1GB are generally good, going over 1 or 2GB means you have a really big dataset and/or a lot of memory available per core,
```

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+- Upper bound: Avoid too very large task graphs. More than 10,000 or 100,000 chunks will perform poorly for this reason.
```
guillaumeeb:
```suggestion
- Upper bound: Avoid too large task graphs. More than 10,000 or 100,000 chunks may start to perform poorly.
```

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+- Lower bound: To get the advantage of parallization, you need the number of chunks to at least equal the number of workers available (or better, the number of workers times 2). Otherwise, if the number of chunks is less than the number of workers, some workers will stay idle.
```
guillaumeeb:
```suggestion
- Lower bound: To get the advantage of parallelization, you need the number of chunks to at least equal the number of worker cores available (or better, the number of worker cores times 2). Otherwise, some workers will stay idle.
```

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+```python
+rechunked_array = original_array.rechunk(new_shape)
+```
+
+**Warning:** Rechunking Dask arrays comes at a cost.
+- The Dask graph must be rearranged to accomodate the new chunk structure. This happens immediately, and will block any other interaction with python until Dask has rearranged the task graph.
+- This also inserts new tasks into the Dask graph. At compute time, there are now more tasks to execute.
+
+For these reasons, it is best to avoid rechunking. However, sometimes the data is stored on disk is not well aligned and rechunking may be necessary.
+For an example of this, here is Draga Doncila Pop [talking about chunk alignment](https://youtu.be/10Ws59NGDaE?t=833) with satellite image data.
```
guillaumeeb: Should we introduce [rechunker](https://github.com/pangeo-data/rechunker), which has been built to overcome the complexity of rechunking large dask arrays in memory?

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Too small is a problem
```
guillaumeeb, line 57: I'll would put this part and the one after just below the "How do I nkow what chunks my array has?" part, or even just before.

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+
+If array chunks are too small, it's inefficient. Why is this?
+
+Using Dask introduces some amount of overhead for each task in your computation.
+This overhead is the reason the Dask best practices advise you to [avoid too-large graphs](https://docs.dask.org/en/latest/best-practices.html#avoid-very-large-graphs).
+This is because if the amount of actual work done by each task is very tiny, then the percentage of overhead time vs useful work time is not good.
+
+Typically, the Dask scheduler takes 1 millisecond to coordinate a single task. That means we want the computation time for each task to be comparitively larger, eg: seconds instead of milliseconds.
```
guillaumeeb, line 65: This is already said above. But I would remove the other one and keep this one.

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+- [Dask dashboard documentation](http://distributed.dask.org/en/latest/diagnosing-performance.html)
+
+We recommend always having the dashboard up when you're working with Dask.
+It's a fantastic way to get a sense of what's working well, or poorly, so you can make adjustments.
+
+Bad signs to watch out for include:
+- Lots of white space in the task stream plot is a bad sign. White space means nothing is happening. Chunks may be too small.
+- Lots and lots of red in the task stream plot is a bad sign. Red means worker communication. Dask workers need some communication, but if they are doing almost nothing except communication then there is not much productive work going on.
+- On the worker memory plot, watch out for orange bars which are a sign you are getting close to the memory limit. Chunks may be too big.
+- On the worker memory plot, watch out for grey bars which mean data is being spilled to disk. Chunks may be too big.
```
guillaumeeb, line 129: Maybe sum-up the last two points as already explained in previous chapter?

_posts/2021-11-02-choosing-dask-chunk-sizes.md
```diff
@@ -0,0 +1,157 @@
...
+## Too big is also a problem
+
+If the Dask array chunks are too big, this is also bad. Why?
+Chunks that are too large are bad because then you are likely to run out of working memory.
+You may see out of memory errors happening, or you might see performance decrease substantially as data spills to disk.
+
+When too much data is loaded in memory on too few workers, Dask will try to spill data to disk instead of crashing.
+Spilling data to disk makes things run very slowly, because all the extra read/write operations to disk. Things don't just get a little bit slower, they get a LOT slower, so it's smart to watch out for this.
+
+To avoid data being spilled to disk, watch the **worker memory plot** on the dask dashboard.
```
guillaumeeb:
```suggestion
To avoid data being spilled to disk, watch the **worker memory plot** on the Dask dashboard.
```

guillaumeeb (commented):
Some typos and suggestions, I really like all the content here!

_posts/2021-11-02-choosing-dask-chunk-sizes.md
GenevieveBuckley: Much clearer! Thank you very much

_posts/2021-11-02-choosing-dask-chunk-sizes.md
GenevieveBuckley: Oh yes, I think that would be good to include.

_posts/2021-11-02-choosing-dask-chunk-sizes.md
GenevieveBuckley: So perhaps this should say instead "chunk sizes between 100MB and 1GB are generally good"

_posts/2021-11-02-choosing-dask-chunk-sizes.md
GenevieveBuckley: Probably better to keep this and remove the other one. We can link to this from that other section.

_posts/2021-11-02-choosing-dask-chunk-sizes.md
GenevieveBuckley: Added:
> The [rechunker](https://rechunker.readthedocs.io/en/latest/) library can be useful in these situations:
>
> > Rechunker takes an input array (or group of arrays) stored in a persistent storage device (such as a filesystem or a cloud storage bucket) and writes out an array (or group of arrays) with the same data, but different chunking scheme, to a new location. Rechunker is designed to be used within a parallel execution framework such as Dask.

_posts/2021-11-02-choosing-dask-chunk-sizes.md
GenevieveBuckley: I agree that this is important to keep here (and now with the rearranging of sections, this part comes first in the article). 

But I don't want to remove it from the rules of thumb list for two reasons:
1 It is a very clear and specific guideline so I think it belongs, and
2. Most people will assume the "rough rules of thumb" list is a good summary of the takeaways from this article, so if we omit it from the list people will likely forget or overlook this point.
I think a small amount of repetition is ok in this case.

_posts/2021-11-02-choosing-dask-chunk-sizes.md
GenevieveBuckley: Upon further review, my opinion is the same as above. Now that we have rearranged the order of sections, I think the small amount of repetition here is helpful to reinforce the key points.

_posts/2021-11-02-choosing-dask-chunk-sizes.md
GenevieveBuckley: Section order rearranged. I think that is better, thank you

GenevieveBuckley:
Ok I think we've got all of the comments addressed now, so I'm merging this.

When I started writing this I expected the content to be a bit thin, with not much new beyond the twitter thread. But with the excellent suggestions from the review feedback, I think it's a lot better, with improvements to content and better logical ordering. So thank you very much @guillaumeeb and @jacobtomlinson for your suggestions, it's made this article much stronger.

guillaumeeb:
Well done @GenevieveBuckley! 👍

PR #121, opened 2021-11-03
Render image at a reasonable size for the choosing chunk sizes blogpost

GenevieveBuckley:
For some reason https://github.com/dask/dask-blog/pull/120 has fixed 2 of the 3 problems, but not the last one. This PR fixes the last image size problem, by specifying the width and height in the html tags.

Related to:
- https://github.com/dask/dask-blog/pull/119
- https://github.com/dask/dask-blog/pull/120

And yes, this time I have checked the html build with jekyll locally. So it *should* finally look nice now!


GenevieveBuckley:
It does look nice now :tada: 
https://blog.dask.org/2021/11/02/choosing-dask-chunk-sizes

PR #128, opened 2021-12-10
Resize too-large images in blogposts

### LINKED ISSUE #127: Resize images
The images in [this blogpost](https://blog.dask.org/2021/03/19/image-segmentation) appear way too large. The size should be reduced so they fit comfortably. This might be happening on other posts (especially other posts of mine) too.

There are two possible approaches to fix this:
1. Use html syntax instead of markdown `alt text` with a maximum width of 700 pixels (this seems to be roughly the same `
2. Reduce the resolution of the image files and commit those with git (additional benefit, slightly smaller repository size).

GenevieveBuckley:
Closes https://github.com/dask/dask-blog/issues/127

There were a few blogposts where the images displayed much larger than expected. This PR makes the image sizes more sensible in three blogposts:
- image segmentation
- high level graphs update, and 
- the 2021 Dask user survey results

### DIFF
--- a/_posts/2021-03-19-image-segmentation.md
+++ b/_posts/2021-03-19-image-segmentation.md
@@ -154,7 +154,7 @@ viewer.add_image(absolute_threshold)
 viewer.add_image(images, contrast_limits=[0, 2000])
 ```
 
-![Absolute threshold napari screenshot](/images/2021-image-segmentation/napari-absolute-threshold.png)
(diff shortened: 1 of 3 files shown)


GenevieveBuckley:
I've previewed the content changes locally, and am going to go ahead and merge this here. It's not a controversial change that requires a code review.

PR #134, opened 2022-04-27
Add categories to RSS feed

### LINKED ISSUE #133: Add tags metadata to blogposts
Hi Folks, currently it looks like we currently track tags in each of the markdown files for each blogpost, but we don't expose these in the HTML or XML/RSS/Atom feeds.  

We're currently trying to migrate over our blog to a new platform and it would be useful to have these in the RSS/Atom feeds, even if they're not visible anywhere.  

Does anyone have time to add this?  At minimum it would probably mean looking up what approach people use by default, and then editing https://github.com/dask/dask-blog/blob/gh-pages/atom.xml or the layouts directory to include tags somewhere.  (Grepping for tags yields some interesting results)

@jacobtomlinson is this easy for you or someone around you?

jacobtomlinson:
closes #133


mrocklin:
Thanks @jacobtomlinson !

PR #135, opened 2022-06-24
Add link to "High level performance of Pandas, Dask, Spark, and Arrow"

scharlottej13:
@jacobtomlinson @mrocklin I was wondering if it would be OK to add a link to a recent Steppingblocks interview with Coiled on their switch from Spark to Dask to your blog post. I started it at "Why we switched from Spark to Dask" since that section of the video seems most relevant. Full video [here](https://youtu.be/jR0Y7NqKJs8).

This is tangentially related to https://github.com/dask/community/issues/259.

cc @adireske

### DIFF
--- a/_includes/themes/twitter/page.html
+++ b/_includes/themes/twitter/page.html
@@ -1,4 +1,5 @@
 <div class="page-header">
+  <!-- <h1 id="page"> -->
   <h1>
     {{ page.title }} {% if page.tagline %} <small>{{ page.tagline }}</small>{%
     endif %}
--- a/assets/themes/twitter/css/style.css
+++ b/assets/themes/twitter/css/style.css
@@ -1,5 +1,96 @@
 /* Override some defaults */
 
+@import url("https://fonts.googleapis.com/css2?family=Inter:wght@400;700");
+@import url("https://fonts.googleapis.com/css2?family=Inconsolata:wght@400;700");
+
+:root {
+    /* Main colors */
+    --yellow: #ffc11e;
+    --white: #fff;
+    --light-grey: #f6f6f6;
+    --grey: #33363d;
+    --dark-grey: #262326;
+    --black: #080815;
+    --red: #ef1161;
+    --pst-color-navbar-link: var(--yellow);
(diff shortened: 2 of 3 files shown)


jacobtomlinson (commented):
I totally agree that it makes sense to add a link here to a video about folks switching from Spark to Dask. And the Steppingblocks video seems great. I'm generally in favour of connecting the two pieces of content.

It does feel a little strange to retroactively add links within the Dask blog to Coiled properties like this though. Especially when the blog post has a `This work is supported by Anaconda Inc` notice at the top. Updating old content feels a bit different compared to new content that is created by folks at Coiled like we are discussing in https://github.com/dask/community/issues/259.

I wonder if one way we could do this without it feeling like a Coiled marketing push would be to snip the three minutes from the YouTube video and put it on the Dask YouTube channel as a separate video. That short video could then call out to the Coiled video for folks to go and watch the rest of the interview. Then we link from the blog to the Dask channel video instead.

Adding that extra hop might help folks switch mindset more cleanly between "This is Dask content sponsored by Anaconda" > "This is Dask content sponsored by Coiled" > "This is Coiled content about Dask".

scharlottej13:
Thanks @jacobtomlinson for taking the time to clearly explain this. I also read through [your comment](https://github.com/dask/community/issues/259#issuecomment-1167263925) on https://github.com/dask/community/issues/259 which adds some nice context.

I updated the PR to link instead to a clip of the Steppingblocks interview on the Dask youtube channel on the [Dask Use Cases playlist](https://youtube.com/playlist?list=PLJ0vO2F_f6OAE1xiEUE7DwMFWbdLCbN3P). I think this adds the extra hop you mentioned, but looking forward to hearing your thoughts. We can also edit an even shorter clip, as you mentioned above.

jacobtomlinson (approved):
Apologies for the delay in coming back here. That's exactly the kind of hop I had intended. thanks for humouring me here.

scharlottej13:
> Apologies for the delay in coming back here. That's exactly the kind of hop I had intended. thanks for humouring me here.

No problem! I think the extra hop makes a lot of sense

PR #142, opened 2022-08-05
Fix broken youtube link in dask-down-under post

GenevieveBuckley:
This youtube link is broken, here is a fix.

It looks like the original youtube video was deleted, so the same content could be re-uploaded with a different title slide (probably connected to the recent logo re-branding).


jacobtomlinson (approved):
Thanks @GenevieveBuckley!

PR #150, opened 2022-10-07
Article: Dask Operator Generally Available

jacobtomlinson:
With the `2022.10.0` release of `dask-kubernetes` the new operator is now generally available.

I would like to publish this blog post which outlines a lot of the reasoning why we rebuilt things from scratch, how that made things simpler and more maintainable and to outline the new features it adds.


quasiben:
This is a really great post!  There is a lot of content here and summary (either at the end or a TLDR would be helpful).  Something like: 

With the new operator you get:
- Cascaded deletions for proper teardown 
- Multiple worker groups for heterogenous/tagged deployments
- DaskJob: running dask workloads with K8s batched job infrastructure
- ...

mrocklin:
Lots of good engagement on this on social media.  Nice work Jacob!

PR #154, opened 2022-12-06
Add resource links to 2021-12-01-mosaic-fusion.md

GenevieveBuckley:
This PR adds several additional resource links to the "Also see" section of the mosaic image fusion blogpost.

I recently had a [recent conversation](https://twitter.com/beniroquai/status/1599534562715242496) with someone who read this blogpost and asked about other tools. That discussion resulted in links to several software packages I was unfamiliar with. I think it would be a good idea to include them in the further reading section, for any future readers also interested in this area.


pavithraes:
@GenevieveBuckley Thank you for adding these links, it looks good to me!

Would you mind updating the bullet point to be defined with `-` instead of `*` to make the linter happy? We can merge after. :)

GenevieveBuckley:
> Would you mind updating the bullet point to be defined with `-` instead of `*` to make the linter happy? We can merge after. :)

Of course, done!

jacobtomlinson (approved):
Thanks @GenevieveBuckley!

I've pushed a commit that fixes the unhappy linter. For future reference you can [set this up automatically with `pre-commit`](https://github.com/dask/dask-blog#formatting).

PR #155, opened 2023-01-10
Update outdated coiled blog links

scharlottej13:
Found some outdated coiled blog links and have updated them. Thanks!

cc @GenevieveBuckley @jacobtomlinson


jsignell (approved):
Thanks Sarah!

PR #162, opened 2023-04-12
Fix broken post title and build instructions

jacobtomlinson:
Looks like Jekyll wasn't happy about the `:` in the title. 

Make a small tweak to the build instructions after reproducing locally.


jacobtomlinson:
I've also enabled build previews with Netlify so we can catch these things more easily.

PR #165, opened 2023-05-02
Set scheduler GPU article back to draft state

mrocklin:
This conveys wrong information currently, which is causing some confusion.

cc @jacobtomlinson would you be ok with this as a short term solution?


jacobtomlinson:
I think this is a bit of an extreme move. It's not incorrect, this is our agreed best practice. Would prefer to soften up the language a little to make it clear that folks can go off-road.

mrocklin:
> this is our agreed best practice

I'm not sure that I agree with this.  If someone wants to use PyTorch and Dask I think I wouldn't tell them "you need to have a GPU on your client and scheduler machines" I would probably point them to the pytorch example that works great from a mac. 

This article says that the thing I think is best practice isn't possible.  This article is wrong. 

> Would prefer to soften up the language a little to make it clear that folks can go off-road.

Sounds great.  Is this something you have time to do or is this something that you're expecting me to do?  My guess is that we're both busy.  That's fine.  While we're both busy, I propose that we unpublish the article that gives incorrect advice.

jacobtomlinson:
Closing in favour of #166

PR #151, opened 2022-11-18
Cross-post queueing blog post

scharlottej13:
cross-post @gjoseph92's queueing blog post with canonical link to coiled blog, see https://github.com/dask/community/issues/291#issuecomment-1317737195

also updated `_config.yml` per a deprecation warning.

@gjoseph92 I copied this from github.com/coiled/blog-drafts. I noticed some differences between this version + https://www.coiled.io/blog/reducing-dask-memory-usage and tried to update accordingly, but I may have missed something. Also had some trouble adding in your figure captions (for some reason using the figcaption html wasn't rendering) so I used some italics and section lines instead.

cc @mrocklin @jacobtomlinson @jrbourbeau

PR #167, opened 2023-05-10
Cross post Dask upstream testing blog post

scharlottej13:
@jrbourbeau wrote this for the Coiled blog, but the content is Dask-focused so I think it would be a good fit for cross-posting on blog.dask.org (see https://github.com/dask/community/issues/291#issuecomment-1317749720 for some discussion around this, noting we did something similar in https://github.com/dask/dask-blog/pull/151).

I made no content changes from the [original post](https://blog.coiled.io/blog/dask-upstream-testing.html).

cc @jacobtomlinson


jacobtomlinson (approved):
Fantastic thank you!

PR #168, opened 2023-05-17
Cross-post Dask observability post

scharlottej13:
@hendrikmakait wrote this for the Coiled blog, but the content is Dask-focused so I think it would be a good fit for cross-posting on blog.dask.org (see https://github.com/dask/community/issues/291#issuecomment-1317749720 for some discussion around this, noting we did something similar in https://github.com/dask/dask-blog/pull/151).

I made no content changes from the [original post](https://blog.coiled.io/blog/observability-with-dask.html).

cc @jacobtomlinson


_posts/2023-05-16-observability-with-dask.md
```diff
@@ -0,0 +1,253 @@
...
+{% include JB/setup %}
+
+_Original version of this post appears on [blog.coiled.io](https://blog.coiled.io/blog/observability-with-dask.html?utm_source=dask-blog&utm_medium=observability-with-dask)_
+
+**Debugging is hard.
+Distributed debugging is hell.**
+
+When dealing with unexpected issues in a distributed system, you need to understand what and _why_ it happened, how interactions between individual pieces contributed to the problems, and how to avoid them in the future.
+In other words, you need **observability**.
+This article explains what observability is, how Dask implements it, what pain points remain, and how Coiled helps you overcome these.
```
jacobtomlinson, line 29: This line especially sets the reader up for a sales pitch.

jacobtomlinson (changes requested):
Thanks @scharlottej13. I love the general intent of this post. However, the post interweaves Dask features with Coiled features throughout so I'm not sure it's the best fit for the Dask community blog without some heavy modification. It feels like too much of a Coiled sales pitch in its current format.

I see three options here:
1. Rework the post to remove the Coiled-specific content and instead point readers to the Coiled blog version at the end for further reading. I think it's fine to direct folks to Coiled to learn about how Coiled solves this problem after discussing the OSS options and their pros/cons.
2. Add a "Sponsored Post" banner at the top to make it clear that this is a Coiled advert. I'm not sure if we want to go down this road in general, but I know it's been discussed before so maybe the Dask Owners need to chat about this.
3. We don't cross-post this one

If we want to go with option 1 let me know and I'll go through and review it through that lens. Also happy to chat about the other options, maybe a call would be easier to get into the nuance here?

hendrikmakait:
I agree with @jacobtomlinson, in its current form the post does not seem to be appropriate for the Dark blog. I'd personally be up for modifying in into a format that feels less sponsored, but I could use an independent sparring partner to make this a quick piece of work. @jacobtomlinson, would you want to collaborate on this?

jacobtomlinson:
Very happy to help. I don't have capacity this week but maybe next week we could chat about this?

scharlottej13:
Thanks @jacobtomlinson and @hendrikmakait! I'll close this then (and open a new PR for the css changes I suggested, which is what I should have done in the first place 😄)

PR #169, opened 2023-05-17
Cross post p2p shuffling

scharlottej13:
@hendrikmakait wrote this for the Coiled blog, but the content is Dask-focused so I think it would be a good fit for cross-posting on blog.dask.org (see https://github.com/dask/community/issues/291#issuecomment-1317749720 for some discussion around this, noting we did something similar in https://github.com/dask/dask-blog/pull/151).

I made no content changes from the [original post](https://blog.coiled.io/blog/shuffling-large-data-at-constant-memory.html).

cc @jacobtomlinson

PR #170, opened 2023-05-18
Add CSS styling option for centering figures with captions

scharlottej13:
Pulled out of #168, will let you do something like:

```
<figure class="align-center">
<img src="/images/my_image.png"/>
<figcaption>
<em>my caption</em>
</figcaption>
</figure>
```
As done in #169


jacobtomlinson (approved):
Thanks!

PR #176, opened 2024-11-20
Cross post dask detrending from xarray blog

scharlottej13:
Cross-posting @phofl's [xarray blog post](https://xarray.dev/blog/dask-detrending).


GenevieveBuckley:
According to the logs, the build is failing because it needs ruby >=3.0.0 for some of the dependencies to install correctly.

We could edit this line of the gemfile to try and fix that. (I think currently `RUBY_VERSION` is not actually defined anywhere, and so it's just making a wild guess and trying to install ruby 2.7.2 for us)
https://github.com/dask/dask-blog/blob/0d3d6e1ed6194cd598fed8d2e48a0025fb68b343/Gemfile#L2

jacobtomlinson:
Thanks for cross-posting here @scharlottej13. And thanks for investigating the build failure @GenevieveBuckley.

I've pulled in the build fix from #177, so things should build successfully now.

jacobtomlinson (changes requested):
Looks like some of the figures are rendering a bit big. Could you take a look at set a maximum width @scharlottej13?

You might need to use HTML `` or something along those lines instead of the markdown syntax for images.

scharlottej13:
Thanks @jacobtomlinson and @GenevieveBuckley!

I think these image sizes look much more reasonable

jrbourbeau (approved):
Thanks @scharlottej13 

(And thanks @GenevieveBuckley @jacobtomlinson for reviewing)

PR #177, opened 2024-11-21
Fix build failures

jacobtomlinson:
Exploring fixing the build issues seen in #176


jacobtomlinson:
Build is passing with an explicit `.ruby-version`.

jrbourbeau (commented):
Thanks @jacobtomlinson

